PR #66, opened 2026-06-14
[codex] Harden batch source index workflow

hu-border-collie:
## Summary

This PR hardens the batch/source-index workflow based on issues observed during a real Ren'Py translation preparation run.

Changes:
- Allow `thinking_level` to be explicitly disabled so Gemini requests omit `thinking_config` instead of showing misleading default-thinking warnings.
- Treat transient embedding transport failures, including SSL EOF/read/connect errors, as retryable and rotate API keys when available.
- Persist source-index bootstrap embeddings incrementally in small batches and print progress, so a transient failure no longer discards all completed embedding work.
- Print source-index retrieval progress during batch build, making long build steps and retry loops visible by chunk.
- Relax the keyword extraction schema by removing the nested `category.enum`, while preserving category normalization in parser logic.

## Root Cause

The workflow exposed three practical issues:
- Gemini rejected the keyword schema when `category.enum` appeared inside the nested candidate schema.
- Embedding calls could fail with transient service/transport errors without retrying or preserving completed source-index progress.
- Empty `thinking_level` config correctly omitted `thinking_config`, but the warning/banner still implied default thinking behavior.

## Validation

- `python -c "import ast, pathlib; [ast.parse(pathlib.Path(p).read_text(encoding='utf-8')) for p in ['gemini_translate_batch.py','translation_core.py','tests/test_batch_rag.py','tests/test_translation_core.py']]; print('syntax ok')"`
- `python -m unittest discover -s tests -p "test_*.py" -q`
- Real workflow smoke: `sync-keywords`, `build-keywords`, `bootstrap-source-index`, and full `build --skip-prepare` completed on a local Ren'Py project; generated requests omitted `thinking_config` and source-index retrieval completed for all chunks.

### DIFF
--- a/gemini_translate_batch.py
+++ b/gemini_translate_batch.py
@@ -67,7 +67,10 @@ def initialize_batch_logging():
     if isinstance(sys.stdout, DualLogger):
         return
     ensure_batch_dirs()
-    sys.stdout = DualLogger(CONSOLE_LOG)
+    try:
+        sys.stdout = DualLogger(CONSOLE_LOG)
+    except OSError as exc:
+        print(f'Warning: Could not open console log {CONSOLE_LOG}: {exc}')
 
 BATCH_MODEL = 'gemini-3.1-flash-lite'
 BATCH_TARGET_SIZE = 60
@@ -1424,6 +1427,17 @@ def copy_split_context_metadata(source_manifest, part_manifest, part_chunks):
             max_context_chars=story_settings.get('max_context_chars'),
         )
 
+    for key in (
+        'source_index_enabled',
+        'source_index_store_path',
+        'source_index_settings',
+    ):
+        if key in source_manifest:
+            part_manifest[key] = source_manifest[key]
+
+    if source_manifest.get('source_index_enabled'):
+        part_manifest['source_index_summary'] = summarize_batch_source_index(part_chunks)
+
 
 def split_chunks_and_lines(chunks, request_lines, max_chunks=0, max_items=0):
     groups = []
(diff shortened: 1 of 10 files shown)


hu-border-collie:
Follow-up from the real Game_GloryHounds batch run:

The full translation batch built successfully, but `check` reported 184 row-level `INVALID_ARGUMENT` errors. The successful rows were the smallest chunks, while chunks with larger item counts failed before generation. Google’s structured-output docs call out schema complexity as a rejection cause, and our per-chunk schema was making `id` an enum of every long item id in that chunk.

This update removes the long `id.enum` constraint from translation and revision response schemas while keeping `id` as a required string. Safety still comes from the existing manifest/result validation in `check` and `apply`, which reject missing, unknown, duplicate, mismatched, or source-mismatched items before writing files.

Verification: `python -m unittest discover -s tests -p "test_*.py" -q` passes locally.

hu-border-collie:
Second real-run follow-up from `Game_GloryHounds` at `batch.chunk_size=40`:

- New package: `20260615_120850_Game_GloryHounds`
- Result rows matched expected chunks: `285/285`
- Row-level API errors dropped to `0`, confirming the schema simplification fixed the previous structured-output rejection.
- Initial local validation had 141 failures, but many were false positives for short preserve terms adjacent to Chinese text, e.g. `Lou笑了` and `Max做饭`.

This update switches short alphabetic preserve-term validation from Unicode `\b` word boundaries to ASCII word boundaries, so Chinese-adjacent preserved names are accepted without matching fragments inside longer English words. It also strengthens translation/revision prompts to require exact verbatim preserve-term substrings, including honorifics, apostrophes, numbers, and spacing.

Rechecking the downloaded batch with the updated validator reduced failures from 141 to 70. Remaining failures are real batch quality issues: one duplicate/misaligned chunk, three missing ids, and strict preserve-term/no-Chinese validation failures. Verification: full unittest discovery passes locally.

hu-border-collie:
Added the retry/merge workflow requested after the `Game_GloryHounds` `chunk_size=40` run:

- New command: `build-retry <parent-manifest>`
  - Rechecks the parent results, finds unsafe/warn chunks, and creates a local retry package under `retry_parts/`.
  - Regenerates request rows from the original chunk data using the current script prompt/schema, instead of copying stale request JSONL.
  - Carries over context metadata including Source Index settings/summary.
- New command: `merge-retry <parent-manifest> <retry-manifest>`
  - Validates retry chunks match the parent target shape.
  - Replaces only retry chunk result rows in the parent result set.
  - Writes a new `results.merged_*.jsonl`, updates the parent manifest, clears stale check metadata, and requires a fresh `check` before `apply`.
- Console logging now warns and continues if the shared console log file cannot be opened.

Real package smoke test:

`build-retry` on `20260615_120850_Game_GloryHounds` produced `retry_parts/20260615_123511_retry` with 56 chunks / 2240 items, matching the affected-chunk analysis. The retry requests have `id_enum_rows=0` and all include the hardened preserve-term prompt.

Verification: `python -m unittest discover -s tests -p "test_*.py" -q` passes locally.

PR #245, opened 2026-07-21
feat: require explicit workspace and stop silent path discovery

hu-border-collie:
## Summary

- **Workspace**: default unset; must choose in GUI or set `workspace_root` / `--workspace` (no tool-parent default).
- **Tool paths**: `api_keys.json` and batch logs stay under the tool root only (no `../data` or parent logs fallback).
- **game_root**: empty until configured; no `ROOT_DIR/..` placeholder project.
- **Ren'Py SDK**: prepare no longer auto-scans nearby dirs; GUI adds **查找 SDK** / **浏览…** for explicit setup.
- **Registry**: open-on-show Game_* auto-scan remains **off** by default; users can enable the checkbox.

## Test plan

- [ ] Fresh config without `workspace_root`: project list shows empty state; choose workspace works
- [ ] CLI `games_registry.py show` without workspace fails with clear message; with `--workspace` works
- [ ] No `game_root`: doctor/translate need an explicit project path (no silent parent)
- [ ] SDK field empty: prepare does not pick nearby SDK; **查找 SDK** finds and fills path
- [ ] Auto-scan checkbox off by default; enabling scans only inside workspace
- [ ] `pytest tests/test_games_registry.py tests/test_gui_project_state.py tests/test_translator_runtime.py tests/test_gui_games_registry_view.py tests/test_gui_app_config.py -q`

## Summary by CodeRabbit

* **新功能**
  * GUI 新增/调整工作区管理入口为「设置 → 项目列表」，首次接入需先选择工作区根；未选择时不支持扫描/导入/切换，页面呈空状态。
  * 当“打开分区时自动扫描新项目”为默认关闭。
  * Ren’Py SDK 改为需显式指定：通过「查找 SDK（主动扫描）」或「浏览…（手动选择）」，留空则不自动扫描。
* **改进**
  * 工作区不再隐式推断：CLI 不设置时项目列表为空且相关命令以退出码 2 报错；CLI 示例优先显式传入/配置工作区。
  * 工具数据目录布局更明确：日志与密钥等统一落在工具目录下，运行时不再回退到当前目录。


translator_runtime.py
```diff
@@ -1414,7 +1516,9 @@ def load_translator_settings():
         else:
             BASE_DIR = original_root
     else:
-        BASE_DIR = _canonical_abs_path(os.path.join(ROOT_DIR, ".."))
+        # No configured game_root: leave empty (do not invent tool parent as project).
+        BASE_DIR = ""
```
hu-border-collie, line 1523: **[bug]** Leaving `BASE_DIR` empty is the right policy, but several prepare/bootstrap helpers still build paths with `os.path.join(BASE_DIR, …)` / `os.path.abspath(BASE_DIR)` when `BASE_DIR` is `""`. On Windows (and POSIX), that quietly binds to the process CWD: e.g. `resolve_work_dir()` becomes `<cwd>/work`, `resolve_original_game_dir()` can pick `<cwd>/original/game`, `_resolve_prepare_launcher()` globs `*.py`/`*.exe` in CWD, and relative `prepare.source_game_dir` / launcher paths resolve against CWD. A CLI `prepare`/`bootstrap-work`/`doctor` run without `game_root` can therefore invent or touch a project under the current directory — the same class of silent path discovery this PR aims to stop (only the old “tool parent” default was removed).

**Suggestion:** Treat empty `BASE_DIR` as a hard precondition failure before any prepare/bootstrap/launcher/path resolution (fail with a clear “set game_root / GAME_ROOT” error). Guard helpers such as `resolve_work_dir`, `resolve_project_root`, `_resolve_prepare_launcher`, `_resolve_path`, and `run_prepare_steps` so they never call `abspath`/`join`/`glob` on an empty base. Add a regression test that loads settings with no `game_root` and asserts bootstrap/prepare path helpers do not create or target `<cwd>/work`.

translator_runtime.py
```diff
@@ -1468,17 +1579,22 @@ def load_translator_settings():
...
-        (BASE_DIR, ROOT_DIR, TOOL_DIR),
-    )
-    if _is_renpy_sdk_dir(resolved_renpy_sdk_dir):
-        PREP_RENPY_SDK_DIR = resolved_renpy_sdk_dir
+    if isinstance(renpy_sdk_dir, str) and renpy_sdk_dir.strip():
+        resolved_renpy_sdk_dir = _resolve_preferred_path_from_bases(
+            renpy_sdk_dir,
+            (BASE_DIR, ROOT_DIR, TOOL_DIR),
+        )
+        PREP_RENPY_SDK_DIR = (
```
hu-border-collie: **[suggestion]** When `prepare.renpy_sdk_dir` or `RENPY_SDK_DIR` is set but does not look like an SDK (`renpy.py` missing), `PREP_RENPY_SDK_DIR` is silently forced to `""`. That matches “no auto-discover fallback,” but users get no warning that a configured path was ignored, so doctor/prepare later only reports “not configured” while the settings field still shows the original string from JSON.

**Suggestion:** Keep the no-fallback behavior, but emit a clear warning (CLI print / doctor note) that the configured SDK path was rejected and why, without scanning nearby directories.

games_registry.py
```diff
@@ -107,19 +107,94 @@
...
+    path = Path(config_path) if config_path is not None else translator_config_path()
+    resolved = Path(workspace).expanduser().resolve()
+    data: dict[str, Any] = {}
+    if path.is_file():
+        try:
+            loaded = json.loads(path.read_text(encoding="utf-8-sig") or "{}")
+            if isinstance(loaded, dict):
+                data = loaded
+        except (OSError, json.JSONDecodeError, TypeError, ValueError):
+            data = {}
```
hu-border-collie: **[suggestion]** `save_configured_workspace_root` catches JSON/OS errors and falls back to `data = {}`, then rewrites `translator_config.json` with only `workspace_root`. A corrupt or unreadable config would be clobbered and lose `game_root`, models, prepare settings, etc. The GUI path (`ProjectState._save_workspace_root_to_config`) correctly raises instead; this helper is currently only used in tests but is part of the public API surface added by the PR.

**Suggestion:** On parse/read failure, raise (or refuse to write) rather than overwriting with a one-key object. Mirror `ProjectState._read_json_object` failure semantics.

tests/test_translator_runtime.py
```diff
@@ -709,10 +735,60 @@ def test_prepare_invalid_configured_sdk_falls_back_to_discovery(self):
...
+
+            with (
+                mock.patch.object(runtime, 'TRANSLATOR_CONFIG', str(config_path)),
+                mock.patch.object(runtime, 'ROOT_DIR', str(tool)),
+                mock.patch.object(runtime, 'TOOL_DIR', str(tool)),
                 mock.patch.dict(os.environ, {}, clear=True),
             ):
                 runtime.load_translator_settings()
-                self.assertEqual(runtime.PREP_RENPY_SDK_DIR, str(sdk))
+                self.assertEqual(runtime.BASE_DIR, '')
```
hu-border-collie, line 785: **[suggestion]** New tests assert empty `BASE_DIR` / no SDK auto-discovery on load, which is good, but they do not exercise the new failure modes users will hit after upgrade: CLI/registry without workspace (covered elsewhere), prepare/bootstrap with empty `game_root` must not touch CWD, invalid configured SDK should not silently look like “never set,” and GUI “查找 SDK” search roots when workspace/game are unset.

**Suggestion:** Add focused regression tests for (1) `resolve_work_dir()` / `run_prepare_steps` / bootstrap with empty `BASE_DIR` refusing CWD, (2) invalid `renpy_sdk_dir` remaining non-discovered, and optionally (3) panel/CLI messaging when workspace is unset.

gui_qt/app.py
```diff
@@ -3166,6 +3172,112 @@ def _advanced_setting_row(self, field: SettingField, widget: QWidget) -> QWidget
...
+        from translator_runtime import discover_renpy_sdk_candidates
+
+        game_root = self.state.get_game_root()
+        workspace_root = self.state.get_workspace_root()
+        tool_root = self.state.get_tool_root()
+        candidates = discover_renpy_sdk_candidates(
+            game_root=str(game_root) if game_root is not None else None,
+            tool_root=str(tool_root),
+            workspace_root=str(workspace_root) if workspace_root is not None else None,
+            include_runtime_defaults=True,
```
hu-border-collie: **[nit]** 「查找 SDK」calls `discover_renpy_sdk_candidates(..., include_runtime_defaults=True)`, which also walks runtime `BASE_DIR` parents and the parent of the tool root. The “未找到” dialog only lists project / workspace / tool roots, so the UI under-describes where the scan actually looked (and may still visit tool-parent dirs that the rest of the PR treats as non-defaults).

**Suggestion:** Either pass `include_runtime_defaults=False` and only search the explicit GUI roots (game/workspace/tool, plus intentional parents of those), or expand the roots hint to match the real search set.

games_registry.py
```diff
@@ -107,19 +107,94 @@
...
+    if not path.is_file():
+        return None
+    try:
+        data = json.loads(path.read_text(encoding="utf-8-sig") or "{}")
+    except (OSError, json.JSONDecodeError, TypeError, ValueError):
+        return None
+    raw = data.get("workspace_root")
+    if not isinstance(raw, str) or not raw.strip():
+        return None
+    return Path(raw.strip()).expanduser()
```
hu-border-collie: **[nit]** `load_configured_workspace_root` returns `Path(raw).expanduser()` without resolving/absolutizing. Relative `workspace_root` values in config then depend on process CWD until `require_workspace_root` resolves them, while the GUI always persists absolute paths via `canonical_abs_path`. Mixed CLI/GUI use of a relative config value can point at different directories.

**Suggestion:** Normalize with `.expanduser().resolve()` (or the shared canonical helper) at load time, and/or reject non-absolute values when saving.

hu-border-collie (commented):
审查结论：主线方向正确（显式 workspace / 去掉静默路径发现），但建议先修复空 BASE_DIR 仍落到 CWD 的问题，再合入。详见行内评论。

（GitHub 不允许对自己的 PR 使用 Request changes，故以 Comment 提交。）

translator_runtime.py
```diff
@@ -2545,12 +2721,22 @@ def _slugify(text):
...
+    base_name = os.path.basename(base)
+    if base_name.lower() in {"work", "original"}:
+        parent = os.path.basename(os.path.dirname(base))
         return _slugify(parent or base_name)
-    project_root = resolve_project_root(base_dir)
-    return _slugify(os.path.basename(os.path.normpath(project_root)))
+    return _slugify(base_name)
 
 
 def guess_project_slug():
```
hu-border-collie, line 2742: **[bug]** Runtime now correctly returns "unset" when BASE_DIR is empty (_project_slug_from_base_dir / guess_project_slug), which is necessary after this PR stops inventing a default game root. However, gemini_translate_batch.guess_project_slug still does os.path.basename(os.path.abspath(legacy.BASE_DIR)). For empty BASE_DIR, os.path.abspath("") resolves to the process CWD, so package names / RAG metadata can silently adopt the CWD folder name. Paths such as bootstrap_rag_store(..., seed_jsonl_paths=...) can reach get_rag_store() / get_source_index_store() without a real TL dir, making this reachable after the new empty-BASE_DIR default.

**Suggestion:** Delete or redirect the batch-local guess_project_slug to legacy.guess_project_slug() (or share _project_slug_from_base_dir), and add a regression test that BASE_DIR="" yields "unset" (or equivalent) rather than the CWD basename.

gui_qt/project_state.py
```diff
@@ -391,6 +390,49 @@ def _write_json_object(self, path: Path, data: dict[str, Any]) -> None:
...
+        self._workspace_root = None
+
+    def _load_workspace_root_from_config(self) -> None:
+        if not self.config_path.exists():
+            return
+        try:
+            data = json.loads(self.config_path.read_text(encoding="utf-8-sig") or "{}")
+            raw = data.get("workspace_root")
+            if isinstance(raw, str) and raw.strip():
+                self._workspace_root = Path(canonical_abs_path(raw.strip()))
```
hu-border-collie: **[suggestion]** GUI workspace load uses canonical_abs_path(raw.strip()) without expanduser(), while CLI/registry load (games_registry.load_configured_workspace_root) uses Path(...).expanduser().resolve(). A hand-edited workspace_root of ~/RenPy_Workspace is expanded correctly by CLI but becomes a literal ...\~\RenPy_Workspace path in the GUI.

**Suggestion:** Align loaders—e.g. Path(raw.strip()).expanduser() before canonical_abs_path, or reuse load_configured_workspace_root / one shared helper for both GUI and CLI.

translator_runtime.py
```diff
@@ -2563,15 +2749,22 @@ def get_context_storage_location():
 
 def get_context_storage_root(base_dir=None):
     if CONTEXT_STORAGE_LOCATION == "game":
-        return _canonical_abs_path(os.path.join(resolve_project_root(base_dir), CONTEXT_STORAGE_GAME_DIR_NAME))
+        try:
+            project_root = resolve_project_root(base_dir)
+        except MissingGameRootError:
+            # Game-side storage needs a project; fall back to tool logs root.
```
hu-border-collie: **[suggestion]** When context_storage.location == "game" and game_root is unset, get_context_storage_root silently returns LOG_DIR. get_default_context_store_dir then returns os.path.join(LOG_DIR, store_name) with no project slug (unlike the "tool" location, which uses "unset"). That can mask a missing project configuration and place game-mode stores in a flatter, less isolated layout under tool logs.

**Suggestion:** Prefer raising MissingGameRootError (or returning empty and failing callers) when location is "game" without a project; if a fallback is required, use the same .../store_name/unset layout as tool mode and log a clear warning.

gui_qt/games_registry_panel.py
```diff
@@ -461,21 +509,97 @@ def __init__(
...
+        else:
+            self._workspace_path_edit.clear()
+            self._choose_workspace_btn.setText("选择工作区…")
+        self._workspace_empty_state.setVisible(not has_workspace)
+        self._workspace_body.setVisible(has_workspace)
+        self._switch_btn.setEnabled(False)
+        if not has_workspace:
+            self._apply_interaction_gate()
+
+    def _choose_workspace(self) -> None:
```
hu-border-collie, line 580: **[suggestion]** The new workspace empty-state, "选择/更改工作区…", host persist/revert (MainWindow._on_workspace_changed), and action gating when workspace is unset are central UX/correctness paths for this PR, but tests only cover panel construction kwargs and low-level state/config helpers. There is no coverage that actions stay blocked without a workspace, that a successful choose updates config + table, or that a failed persist reverts the panel.

**Suggestion:** Add focused GUI/unit tests for: empty-state visible when workspace_root is None; _require_workspace blocks discover/ingest/refresh; _on_workspace_changed success and ValueError revert; optional checkbox preference still defaults to off after choose.

gui_qt/app.py
```diff
@@ -3166,6 +3172,114 @@ def _advanced_setting_row(self, field: SettingField, widget: QWidget) -> QWidget
...
+        self.statusBar().showMessage(f"已填入 Ren'Py SDK：{path_text}", 5000)
+
+    def _on_find_renpy_sdk_dir(self, line_edit: QLineEdit) -> None:
+        from translator_runtime import discover_renpy_sdk_candidates
+
+        game_root = self.state.get_game_root()
+        workspace_root = self.state.get_workspace_root()
+        tool_root = self.state.get_tool_root()
+        # Only explicit GUI roots (+ their intentional parents via search helper).
+        # Do not also walk runtime defaults / tool-parent implicit bases.
```
hu-border-collie: **[nit]** Comment says find-SDK should not walk "tool-parent implicit bases", and include_runtime_defaults=False correctly skips runtime BASE/ROOT defaults. Passing tool_root= still always adds the tool parent via renpy_sdk_search_roots (and the no-results dialog even advertises "工具：…（及其上一级）"). The comment overstates the restriction and can confuse future changes that try to "honor" it by stripping parent search incorrectly—or leave parent search in place while thinking it is already disabled.

**Suggestion:** Reword the comment to match behavior: explicit GUI roots only; for tool_root / game_root, intentional parent hops still apply; runtime globals are not mixed in.

gui_qt/games_registry_panel.py
```diff
@@ -924,10 +1064,16 @@ def _apply_interaction_gate(self) -> None:
         """Combine local refresh-busy and host workbench-running gates."""
         # Prefer the explicit UI flag so buttons lock as soon as refresh is
         # requested, even before the QThread reports isRunning().
+        if not self._has_workspace():
+            self._choose_workspace_btn.setEnabled(not self._host_task_running)
```
hu-border-collie: **[nit]** When workspace is unset, _apply_interaction_gate disables the title-row "选择工作区…" button during a host task, but does not disable the empty-state action button (same _choose_workspace handler). Clicks during a running workbench task silently no-op because _choose_workspace returns early without feedback.

**Suggestion:** Also gate the empty-state action (or disable _workspace_empty_state while _host_task_running), and/or show a short status/message when choose is blocked.

hu-border-collie (commented):
## Summary

PR #245 correctly removes silent workspace/game_root defaults (tool-parent discovery), pins tool assets (`api_keys.json`, batch logs) under the package root, makes Ren'Py SDK discovery explicit (GUI browse/find only; prepare no longer auto-scans), and gates registry UI/CLI behind an explicit workspace. Core path-boundary changes in `games_registry.py`, `translator_runtime.py`, and the GUI panel look coherent and are backed by solid unit tests for empty defaults, corrupt-config refusal, CLI exit code 2, and no auto-SDK discovery.

Dominant residual risks are incomplete migration outside the new guards: batch still has a CWD-based `guess_project_slug`, GUI vs CLI workspace loaders disagree on `~` expansion, and the new workspace empty-state / choose-workspace path has little integration-test coverage.

## Issue counts by severity

- bugs: 1
- suggestions: 3
- nits: 2

PR #275, opened 2026-07-28
Improve LiteLLM provider catalog settings

hu-border-collie:
## Summary

- replace the fixed, preselected LiteLLM provider/model controls with an explicit provider → credentials → model workflow
- rank common providers first, keep dynamic providers alphabetized, and support case-insensitive contains search while preserving custom input
- add a versioned user-level provider/model catalog cache with per-provider selections, source/timestamp metadata, stale/version warnings, atomic writes, and sensitive-field sanitization
- prefer native provider model endpoints, fall back to the LiteLLM online subset with a visible warning, and retain old cache data on refresh failure
- isolate credentials by provider, clear unsaved plaintext on every provider switch path, and keep saved credentials out of project configuration and catalog cache
- preserve custom providers/models and make saved `sync.litellm_model` take precedence over GUI history
- update the GUI documentation and regression coverage

## Why

The previous settings page mixed a small built-in provider list, default models, transient online results, runtime configuration, and provider credentials. It could look configured before the user made a choice, lost catalog state on restart, and silently injected defaults when refresh failed.

This change makes each state boundary explicit: project runtime configuration remains in `translator_config.json`, catalog and UI history use a user-level cache, and secrets remain in the operating-system credential manager or provider environment variables.

## User impact

Fresh users now start with empty Provider and model selections and no startup network request. Catalog loading is always explicit, common providers stay easy to reach, loaded providers can be searched by any name fragment, cached results survive restarts, failures do not overwrite a usable cache, and custom/Azure/OpenAI-compatible configurations remain available through manual input and LiteLLM environment settings.

## Validation

- `python -m unittest tests.test_gui_litellm_catalog_cache tests.test_litellm_provider_config tests.test_gui_litellm_worker tests.test_gui_litellm_settings_page tests.test_gui_litellm_settings`
- `python -B tests/run_gui_tests.py -q` — 882 tests passed
- `git diff --check`

Closes #274

## Summary by CodeRabbit

* **New Features**
  * Added a cached LiteLLM provider/model selection workflow with an editable provider picker and snapshot-based model loading.
  * Extended supported providers to include **Gemini** and improved dynamic provider discovery from the remote catalog.
* **Documentation**
  * Refined GUI LiteLLM behavior documentation for setup order, credential storage/precedence, and caching/sync boundaries.
* **Bug Fixes**
  * Prevented unintended fallback/auto-defaults when online catalog loading fails; added clearer failure messages.
  * Updated LiteLLM install/settings enablement behavior when required components are missing.
* **Tests**
  * Expanded coverage for cache persistence, invalid/corrupt cache handling, and updated GUI expectations.

### DIFF
--- a/gui_qt/app.py
+++ b/gui_qt/app.py
@@ -235,7 +235,13 @@ from .work_bootstrap_report import (
 from .litellm_worker import (
     LiteLLMConnectionTestWorker,
     LiteLLMModelCatalogWorker,
+    LiteLLMProviderCatalogWorker,
     LiteLLMVersionWorker,
+)
+from .litellm_catalog_cache import (
+    CatalogSnapshot,
+    LiteLLMCatalogCache,
+    catalog_snapshot_warning,
 )
 from .litellm_settings import (
     provider_credential_status,
@@ -251,10 +257,9 @@ from gemini_model_catalog import (
     write_model_catalog_extras,
 )
 from litellm_provider_config import (
-    DEFAULT_MODELS,
-    SUPPORTED_PROVIDERS,
     catalog_source_label,
     installed_litellm_version,
+    provider_display_label,
     version_key,
     ProviderCredentialStoreError,
     delete_provider_api_key,
@@ -457,6 +462,9 @@ _SETTINGS_LAZY_ATTR_TO_PAGE: dict[str, str] = {
     "batch_thinking_combo": "models",
     "sync_backend_combo": "litellm",
     "litellm_provider_combo": "litellm",
+    "litellm_refresh_providers_btn": "litellm",
+    "litellm_clear_provider_btn": "litellm",
+    "litellm_provider_catalog_status_label": "litellm",
     "litellm_model_combo": "litellm",
     "litellm_refresh_models_btn": "litellm",
     "litellm_catalog_status_label": "litellm",
@@ -465,6 +473,7 @@ _SETTINGS_LAZY_ATTR_TO_PAGE: dict[str, str] = {
     "litellm_check_version_btn": "litellm",
     "install_litellm_btn": "litellm",
     "litellm_install_progress": "litellm",
+    "litellm_credentials_box": "litellm",
     "litellm_provider_label": "litellm",
     "litellm_api_key_edit": "litellm",
     "litellm_save_key_btn": "litellm",
@@ -492,6 +501,7 @@ class MainWindow(QMainWindow):
         resources_dir: Path | None = None,
         project_state: ProjectState | None = None,
         bootstrap_config: dict[str, Any] | None = None,
+        litellm_catalog_cache: LiteLLMCatalogCache | None = None,
     ):
         super().__init__()
         self.setWindowTitle("Ren'Py Translation Lab - 图形工作台")
@@ -499,6 +509,7 @@ class MainWindow(QMainWindow):
         self.resize(1180, 780)
 
         self.state = project_state or ProjectState()
+        self._litellm_cache = litellm_catalog_cache or LiteLLMCatalogCache()
         self._diagnostics_context_fingerprint: tuple[object, ...] | None = None
         self._context_library_status_cache: ContextLibraryStatusResult | None = None
         self._context_library_status_cache_at = 0.0
@@ -594,15 +605,14 @@ class MainWindow(QMainWindow):
         self._last_main_tab_index = 0
         self._handling_config_tab_leave = False
         self._task_running = False
+        self._litellm_provider_catalog_worker: LiteLLMProviderCatalogWorker | None = None
         self._litellm_catalog_worker: LiteLLMModelCatalogWorker | None = None
         self._litellm_version_worker: LiteLLMVersionWorker | None = None
         self._litellm_latest_version = ""
         self._litellm_latest_compatible_version = ""
         self._litellm_latest_requires_python = ""
-        self._litellm_catalog_source = ""
         self._litellm_connection_worker: LiteLLMConnectionTestWorker | None = None
         self._font_install_worker: FontInstallWorker | None = None
-        self._litellm_catalog_models: dict[str, tuple[str, ...]] = {}
         self._updating_litellm_provider = False
         # _games_registry_panel is intentionally NOT set here so attribute access
         # triggers __getattr__ lazy materialization of 设置 · 项目列表.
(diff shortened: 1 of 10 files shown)


gui_qt/app.py
```diff
@@ -6560,51 +6740,96 @@ def _refresh_litellm_credential_status(self) -> None:
...
                 saved_message = (
-                    "系统凭据管理器中已保存密钥。"
-                    if load_provider_api_key(status.provider)
+                    "系统凭据管理器中已保存密钥；如同时存在环境变量，请求优先使用此密钥。"
+                    if load_provider_api_key(provider)
                     else "系统凭据管理器中尚未保存密钥。"
                 )
             except ProviderCredentialStoreError as exc:
                 saved_message = str(exc)
         status_label.setText(" ".join(part for part in (saved_message, status.message) if part))
```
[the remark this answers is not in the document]
hu-border-collie: 已在 78e4396 修复：按 Provider 缓存凭据状态，重复刷新不再同步读取 keyring；保存/删除密钥后会立即更新对应缓存。新增回归测试确认多次刷新只调用一次 `load_provider_api_key`。

gui_qt/app.py
```diff
@@ -6560,51 +6740,96 @@ def _refresh_litellm_credential_status(self) -> None:
...
-            else ""
-        )
+        provider = self._litellm_provider_combo_value()
+        api_key_edit = self._settings_widget("litellm_api_key_edit")
+        if api_key_edit is not None:
+            api_key_edit.clear()
+        if not self._loading_config_to_ui:
+            self._save_litellm_cache(
+                lambda: self._litellm_cache.select_provider(provider)
+            )
```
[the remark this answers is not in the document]
hu-border-collie: 已在 78e4396 修复：记录最后实际应用的 Provider，同一 Provider 的 `editingFinished` 直接返回，不再清空未保存 API Key 或自定义模型；程序化恢复/填充时同步该状态，并新增焦点移出回归测试。

tests/test_gui_litellm_catalog_cache.py
```diff
@@ -0,0 +1,126 @@
...
+            self.assertEqual(cache.selected_provider, "")
+            self.assertEqual(cache.providers.values, ())
+            self.assertEqual(cache.models("openai").values, ())
+            self.assertEqual(cache.selected_model("openai"), "")
+
+    def test_catalogs_and_per_provider_selections_survive_reload(self):
+        with tempfile.TemporaryDirectory() as temp_dir:
+            path = Path(temp_dir) / "catalog.json"
+            now = lambda: datetime(2026, 7, 28, 12, 0, tzinfo=timezone.utc)
+            cache = LiteLLMCatalogCache(path, now=now)
```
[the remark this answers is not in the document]
hu-border-collie: 已在 78e4396 修复：将测试中的 lambda 改为局部 `def now()`，消除 Ruff E731。

litellm_provider_config.py
```diff
@@ -140,6 +147,32 @@ def provider_from_model(model: str) -> str:
...
+
+def sort_provider_ids(providers: object) -> tuple[str, ...]:
+    """Place common providers first and sort all remaining ids by name."""
+    if not isinstance(providers, (list, tuple, set, frozenset)):
+        return ()
+    cleaned = {
+        str(provider or "").strip().lower()
+        for provider in providers
+        if str(provider or "").strip()
+    }
```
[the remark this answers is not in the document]
hu-border-collie: 已在 78e4396 修复：`sort_provider_ids` 现在接受 `Collection`，并显式拒绝字符串类标量；新增 `dict_keys` mapping view 覆盖，保持原有归一化与排序行为。

gui_qt/app.py
```diff
@@ -6876,17 +7204,31 @@ def _on_sync_backend_changed(self, _index: int) -> None:
...
+        for name in ("litellm_save_key_btn", "litellm_delete_key_btn"):
             widget = self._settings_widget(name)
             if widget is not None:
-                widget.setEnabled(backend == "litellm" and not installing)
+                widget.setEnabled(credential_enabled)
+        test_button = self._settings_widget("litellm_test_connection_btn")
+        if test_button is not None:
+            test_button.setEnabled(
+                litellm_active
```
hu-border-collie, line 7294: **[bug]** `_on_sync_backend_changed` now enables “测试连接” only when `litellm_active and provider and model`, but `_on_litellm_model_changed` never calls `_on_sync_backend_changed`. After the user selects a provider (button correctly disabled with empty model) and then picks or types a model, the test button can remain disabled indefinitely until some unrelated event re-runs backend gating (provider re-select, catalog worker completion, backend toggle, etc.). Clearing the model has the inverse problem: the button may stay enabled with an empty model. This is a regression versus the old always-on-when-LiteLLM-active gating, introduced by the new model dependency without a model-change re-gate.

**Suggestion:** Call `_on_sync_backend_changed(-1)` at the end of `_on_litellm_model_changed` (or extract a small LiteLLM control-gating helper and invoke it from both provider and model change handlers). Add a settings-page test that selects provider + sets model text and asserts `litellm_test_connection_btn.isEnabled()`.

gui_qt/app.py
```diff
@@ -6560,51 +6762,103 @@ def _refresh_litellm_credential_status(self) -> None:
...
+            if not self._loading_config_to_ui:
+                self._save_litellm_cache(
+                    lambda: self._litellm_cache.select_provider(provider)
+                )
+        provider = provider or self._litellm_provider_combo_value()
+        if provider and model and not self._loading_config_to_ui:
+            self._save_litellm_cache(
+                lambda: self._litellm_cache.select_model(provider, model)
+            )
         self._refresh_litellm_credential_status()
```
hu-border-collie: **[bug]** Model-prefix-driven provider switches in `_on_litellm_model_changed` update the provider combo / clear the API key / persist selected provider, but skip the rest of the provider-change path. In particular they do not: (1) reload that provider’s cached model catalog into the model combo (dropdown still shows the previous provider’s models while credentials target the new prefix), or (2) re-run control gating so ollama credential disablement, clear-provider enablement, and “联网加载模型” state track the effective provider. Example: provider=openai with openai model list visible, user edits model to `ollama/llama3` — key is cleared and applied provider becomes ollama, but credential widgets may remain enabled until another event, and the model list is still OpenAI’s.

**Suggestion:** When the resolved provider changes due to model text, reuse the same path as `_on_litellm_provider_changed` (load `cache.models(provider)`, preserve the typed model as `selected`, refresh catalog status, call `_on_sync_backend_changed`), while still avoiding recursive loops via `_updating_litellm_provider`. Cover with a GUI test for `openai → azure/my-deployment` and `openai → ollama/llama3` side effects.

gui_qt/app.py
```diff
@@ -6560,51 +6762,103 @@ def _refresh_litellm_credential_status(self) -> None:
...
                 provider_combo.setCurrentIndex(index)
                 provider_combo.blockSignals(False)
+            self._applied_litellm_provider = provider
+            if not self._loading_config_to_ui:
+                self._save_litellm_cache(
+                    lambda: self._litellm_cache.select_provider(provider)
+                )
+        provider = provider or self._litellm_provider_combo_value()
+        if provider and model and not self._loading_config_to_ui:
+            self._save_litellm_cache(
```
hu-border-collie: **[suggestion]** `litellm_model_combo.currentTextChanged` persists `select_model` on every keystroke via `_save_litellm_cache` → `atomic_write_json`. Each character rewrite creates a full temp file + replace under the user profile. On Windows this is noisy for AV/indexers, can interleave incomplete model ids into the durable selection, and is unnecessary for a “last selection” cache that only needs to survive process exit.

**Suggestion:** Debounce persistence (e.g. 300–500 ms timer), or write only on `editingFinished` / provider change / successful catalog load / config save. Keep in-memory selection immediate; flush disk asynchronously or on idle.

gui_qt/app.py
```diff
@@ -6510,48 +6564,196 @@ def _selected_sync_backend(self) -> str:
...
+    def _litellm_provider_combo_value(self) -> str:
+        combo = self._settings_widget("litellm_provider_combo")
+        if combo is None:
+            return ""
+        index = combo.currentIndex()
+        if index >= 0 and combo.currentText() == combo.itemText(index):
+            data = str(combo.itemData(index) or "").strip().lower()
+            if data:
+                return data
+        return combo.currentText().strip().lower()
```
hu-border-collie: **[suggestion]** `_litellm_provider_combo_value` returns `itemData` only when `currentText() == itemText(index)`. Free-typed or pasted display labels that do not match the selected index fall through to `currentText().strip().lower()`. That works accidentally for labels that are title-case of the id (`OpenAI` → `openai`) but breaks for labeled entries whose display form is not the id: `Ollama（本地）` → `ollama（本地）`, `Google Gemini` → `google gemini`, `Azure OpenAI` → `azure openai`, `Google Vertex AI` → `google vertex ai`. Those bogus provider ids would then be used for keyring lookups, catalog refresh, and cache keys.

**Suggestion:** Prefer `itemData` whenever `currentIndex() >= 0` and data is set; additionally reverse-map known display labels (`provider_display_label` / combo items) to ids before accepting free text. Placeholder/docs should state that custom entries must be LiteLLM provider ids (e.g. `azure`, `openai`), not marketing names.

gui_qt/app.py
```diff
@@ -6681,6 +6935,51 @@ def _on_litellm_version_checked(
...
+            lambda: self._litellm_cache.update_providers(
+                values,
+                source=source,
+                litellm_version=installed_litellm_version(),
+            )
+        )
+        self._populate_litellm_providers(values, selected=current)
+        self._refresh_litellm_catalog_status()
+        self.statusBar().showMessage(
+            f"已加载 {len(values)} 个 LiteLLM 供应商；未自动选择。",
```
hu-border-collie: **[nit]** After a successful provider catalog load, the status bar always says “未自动选择”, even when `selected=current` restored a previous/custom provider via `_populate_litellm_providers`. The implementation correctly preserves the current selection; the copy does not.

**Suggestion:** Branch the message: if `current` is non-empty after populate, report that N providers were loaded and the current selection was kept; only say “未自动选择” when no provider is selected.

hu-border-collie (commented):
## Summary

This PR successfully replaces fixed preselected LiteLLM provider/model controls with an explicit load → select → credential → model workflow, backed by a versioned user-level catalog cache. Cache design looks solid: atomic writes, schema version rejection, sensitive-field drop on rewrite, stale/version warnings without silent defaults, and credential isolation on provider switch. The dominant residual risks are GUI gating that does not re-run when the model text changes (so “测试连接” can stay disabled after a valid model is chosen), and side effects of treating every `currentTextChanged` as a durable cache write / provider-switch trigger without refreshing the same enablement path used elsewhere.

## Issue counts by severity

- bugs: 2
- suggestions: 2
- nits: 2

## Issues outside the diff

These findings reference lines that are not present in the diff and could not be posted as inline comments:

- **[nit]** litellm_provider_config.py:452 -- `catalog_source_label` still documents a `local` source (“本机 LiteLLM 随包目录（联网失败，可能过时）”), but `LiteLLMModelCatalogWorker` no longer emits `local` fallback models. Dead user-facing branch that can confuse future maintainers about failure behavior.
  - **Suggestion:** Remove the `local` branch (or keep it only if some other caller still emits that token), and rely on the new empty-result + warning dialog path.

PR #278, opened 2026-07-29
Add strict Agent CLI exit codes

hu-border-collie:
## Summary

- add stable machine error classification with retryability and suggested actions
- add `--strict-exit-codes` to the seven core JSON workflow commands
- preserve existing text output and default exit-code behavior
- document the semantic exit-code contract for Agent, Batch, and GUI command references

## Exit code contract

- `0`: success or continue polling
- `2`: CLI usage error
- `3`: completed but needs action, such as `check: warn`
- `4`: blocked by a safety gate or terminal job failure
- `5`: invalid or stale input/configuration/state
- `6`: retryable remote failure

## Validation

- `python -m unittest tests.test_cli_contract tests.test_gemini_translate_batch_cli_contract`
- `python -B tests/run_cli_tests.py -q` — 748 passed, 1 skipped
- `git diff --check`

Part of #276.

### DIFF
--- a/cli_contract.py
+++ b/cli_contract.py
@@ -15,6 +15,130 @@ from typing import Any, TextIO
 
 CLI_SCHEMA_VERSION = 1
 
+EXIT_OK = 0
+EXIT_USAGE = 2
+EXIT_NEEDS_ACTION = 3
+EXIT_BLOCKED = 4
+EXIT_INVALID_STATE = 5
+EXIT_RETRYABLE = 6
+
+
+def classify_error(message: str, *, exception_type: str = "") -> dict[str, Any]:
+    """Classify existing CLI failures without changing their human-readable text."""
+
+    normalized = str(message or "").strip().lower()
+    stale_markers = (
+        "changed after the last check",
+        "stale check",
+        "check fingerprint",
+        "source snapshot",
+        "manifest or results changed",
+    )
+    retryable_markers = (
+        "rate limit",
+        "resource exhausted",
+        "quota",
+        "timed out",
+        "timeout",
+        "temporarily unavailable",
+        "service unavailable",
+    )
+    precondition_markers = (
+        "not succeeded yet",
+        "missing",
+        "not found",
+        "does not exist",
+        "download first",
+        "no job",
+        "api key",
+        "configuration",
+        "config ",
+    )
+    blocked_markers = (
+        "not safe",
+        "safety level",
+        "blocked",
+        "refused",
+        "already applied",
+        "disabled",
+        "cost limit",
+        "max cost",
+    )
+
+    if any(marker in normalized for marker in stale_markers):
+        return {
+            "code": "STALE_STATE",
+            "retryable": False,
+            "suggested_action": "run_check_again",
+            "exit_code": EXIT_INVALID_STATE,
+        }
+    if any(marker in normalized for marker in retryable_markers):
+        return {
+            "code": "REMOTE_RETRYABLE",
+            "retryable": True,
+            "suggested_action": "retry_later",
+            "exit_code": EXIT_RETRYABLE,
+        }
+    if any(marker in normalized for marker in precondition_markers):
+        return {
+            "code": "PRECONDITION_FAILED",
+            "retryable": False,
+            "suggested_action": "inspect_configuration_and_artifacts",
+            "exit_code": EXIT_INVALID_STATE,
+        }
+    if any(marker in normalized for marker in blocked_markers):
+        return {
+            "code": "COMMAND_BLOCKED",
+            "retryable": False,
+            "suggested_action": "inspect_diagnostics",
+            "exit_code": EXIT_BLOCKED,
+        }
+    return {
+        "code": "COMMAND_REFUSED" if exception_type == "SystemExit" else "INTERNAL_ERROR",
+        "retryable": False,
+        "suggested_action": "inspect_diagnostics",
+        "exit_code": EXIT_BLOCKED if exception_type == "SystemExit" else 1,
+    }
+
+
+def strict_exit_code(envelope: Mapping[str, Any]) -> int:
+    """Return the documented semantic exit code for a machine envelope."""
+
+    if not envelope.get("ok"):
+        error = envelope.get("error")
+        if isinstance(error, Mapping):
+            details = error.get("details")
+            if isinstance(details, Mapping):
+                semantic_code = details.get("semantic_exit_code")
+                if isinstance(semantic_code, int):
+                    return semantic_code
+        return EXIT_BLOCKED
+
+    command = str(envelope.get("command") or "")
+    status = str(envelope.get("status") or "").strip().lower()
+    if command == "check":
+        if status == "safe":
+            return EXIT_OK
+        if status == "warn":
+            return EXIT_NEEDS_ACTION
+        if status == "block":
+            return EXIT_BLOCKED
+        return EXIT_INVALID_STATE
+    if command == "doctor" and status == "blocked":
+        return EXIT_BLOCKED
+    if command in {"submit", "status"} and status in {
+        "failed",
+        "cancelled",
+        "canceled",
+        "expired",
+        "job_state_failed",
+        "job_state_cancelled",
+        "job_state_canceled",
+        "job_state_expired",
+    }:
+        return EXIT_BLOCKED
+    return EXIT_OK
+
 
 def _json_compatible(value: Any) -> Any:
     if value is None or isinstance(value, (bool, int, float, str)):
--- a/cli_contract.py
+++ b/cli_contract.py
@@ -26,22 +26,35 @@ EXIT_RETRYABLE = 6
 def classify_error(message: str, *, exception_type: str = "") -> dict[str, Any]:
     """Classify existing CLI failures without changing their human-readable text."""
 
-    normalized = str(message or "").strip().lower()
+    normalized = f"{exception_type} {message or ''}".strip().lower()
     stale_markers = (
         "changed after the last check",
         "stale check",
         "check fingerprint",
         "source snapshot",
         "manifest or results changed",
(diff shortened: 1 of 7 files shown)


cli_contract.py
```diff
@@ -15,6 +15,130 @@
...
+        "stale check",
+        "check fingerprint",
+        "source snapshot",
+        "manifest or results changed",
+    )
+    retryable_markers = (
+        "rate limit",
+        "resource exhausted",
+        "quota",
+        "timed out",
```
hu-border-collie, line 47: **[bug]** `classify_error` retryable markers (`"resource exhausted"`, `"rate limit"`, `"service unavailable"`, `"quota"`, etc.) do not match the failure strings this CLI actually surfaces from Gemini/SDK paths. Existing helpers already treat `429` / `RESOURCE_EXHAUSTED` / `503` / `UNAVAILABLE` / `ConnectTimeout` as retryable (`is_quota_error` / `is_unavailable_error` in `gemini_translate_batch.py`). After `.lower()`, a typical message like `429 RESOURCE_EXHAUSTED` becomes `429 resource_exhausted`, which does **not** contain the space-separated marker `"resource exhausted"`, and also does not contain `"429"`, `"resource_exhausted"`, or bare `"unavailable"`. Submit/status/download failures re-raise those exceptions into `run_machine_command`, so agents following the docs for exit `6` / `REMOTE_RETRYABLE` will usually get `INTERNAL_ERROR` with exit `1` and `retryable=false` instead.

**Suggestion:** Reuse or mirror the same markers as `is_quota_error` / `is_unavailable_error` (include `429`, `resource_exhausted`, `resourceexhausted`, `503`, `unavailable`, connect/read timeout class names). Prefer classifying from structured exception knowledge at the raise site rather than English phrases alone. Add a unit test with a realistic `429 RESOURCE_EXHAUSTED` / `503 UNAVAILABLE` message asserting `REMOTE_RETRYABLE`, `retryable=True`, and `EXIT_RETRYABLE`.

cli_contract.py
```diff
@@ -15,6 +15,130 @@
...
+EXIT_BLOCKED = 4
+EXIT_INVALID_STATE = 5
+EXIT_RETRYABLE = 6
+
+
+def classify_error(message: str, *, exception_type: str = "") -> dict[str, Any]:
+    """Classify existing CLI failures without changing their human-readable text."""
+
+    normalized = str(message or "").strip().lower()
+    stale_markers = (
```
hu-border-collie, line 30: **[bug]** Core apply-preflight refusals that docs map to exit `5` / invalid state are misclassified. In `require_safe_check_for_apply`, `missing_check` uses “Manifest has no valid check summary…” (no `"missing"` / `"not found"` substring) and `stale_check_contract` uses “older check contract…” (no stale marker). Both fall through to `COMMAND_REFUSED` with `EXIT_BLOCKED` (4) and `suggested_action=inspect_diagnostics`, instead of `STALE_STATE` / `PRECONDITION_FAILED` with exit `5` and `run_check_again` / config-artifact guidance. Only the fingerprint-drift message (“changed after the last check”) hits `STALE_STATE`. That is a contract hole on the check→apply safety path agents care about most.

**Suggestion:** Either (preferred) attach structured `reason_code` into the machine error envelope at `fail_apply_preflight` / raise sites and classify from that, or extend markers/tests to cover the exact preflight messages: missing check summary, older check contract, and any other apply-guard strings. Assert exit `5` + stable `error.code` for those cases under `--strict-exit-codes`.

gemini_translate_batch.py
```diff
@@ -11634,39 +11642,63 @@ def run_machine_command(parser, args):
...
-        cli_contract.write_json_envelope(
-            cli_contract.error_envelope(
-                command,
-                code='COMMAND_REFUSED',
-                message=_system_exit_message(exc),
-                details={'exit_code': exc.code if isinstance(exc.code, int) else 1},
-            ),
-            sys.stdout,
+        message = _system_exit_message(exc)
+        classification = cli_contract.classify_error(
```
hu-border-collie: **[suggestion]** Error-code classification runs for every JSON-mode failure, not only when `--strict-exit-codes` is set. Previously `SystemExit` always produced `error.code=COMMAND_REFUSED`; now the same call can yield `STALE_STATE`, `PRECONDITION_FAILED`, `COMMAND_BLOCKED`, etc. Process exit codes stay compatible without the flag, but any agent/script that already branched on `error.code` under plain `--output json` sees a silent schema-v1 contract expansion. Docs emphasize stable `error.code` and “do not parse message,” while classification itself is message parsing—so stability is only as good as current English wording.

**Suggestion:** Call this out explicitly in agent docs/changelog as a JSON envelope change independent of the exit-code flag; consider bumping or documenting additive error codes under schema_version 1. Longer term, emit structured codes at raise sites so `error.code` is not derived from free text.

docs/quickstart_agent.md
```diff
@@ -50,7 +50,21 @@ JSON 模式的 stdout 只包含一个 JSON 文档；banner、进度、warning 
 - `status` 表示业务状态，例如 Batch job state、`safe / warn / block` 或 `applied`；
 - `result` 是命令摘要，`artifacts` 给出 manifest、results、检查报告等产物路径；
 - 命令拒绝执行时 `ok=false`，`error.code` 与 `error.message` 用于程序判断和诊断；
-- 当前默认退出码仍保持兼容；不要仅凭 `check` 的退出码决定是否写回，必须读取 `status`，并且只有 `safe` 才能继续 `apply`。
+- 默认退出码仍保持兼容。Agent 可同时传入 `--output json --strict-exit-codes`，让业务状态映射为稳定退出码；严格模式只与 JSON 输出组合使用。
+- 无论是否启用严格退出码，都必须读取 `status`；只有 `check` 的 `status=safe` 才能继续 `apply`。
+
+严格退出码约定：
+
+| 退出码 | 含义 | 常见场景 |
```
hu-border-collie, line 58: **[suggestion]** The strict-exit table documents `0/2/3/4/5/6` but not exit `1`. `classify_error` still assigns `exit_code: 1` for unclassified non-`SystemExit` failures (`INTERNAL_ERROR`), and `strict_exit_code` will return that value via `details.semantic_exit_code`. Agents implementing only the published table will treat unexpected internal failures as unmapped.

**Suggestion:** Document exit `1` as unexpected/internal error (not retryable by default), and note that `details.semantic_exit_code` is authoritative when present.

tests/test_gemini_translate_batch_cli_contract.py
```diff
@@ -165,9 +171,83 @@ def test_json_mode_turns_system_exit_into_error_envelope(self):
         payload = json.loads(stdout.getvalue())
         self.assertEqual(exit_code, 1)
         self.assertFalse(payload["ok"])
-        self.assertEqual(payload["error"]["code"], "COMMAND_REFUSED")
+        self.assertEqual(payload["error"]["code"], "STALE_STATE")
         self.assertIn("changed after the last check", payload["error"]["message"])
 
+    def test_strict_check_exit_code_reports_needs_action(self):
```
hu-border-collie, line 198: **[suggestion]** End-to-end coverage is thin relative to the contract surface. Tests cover warn→3, stale SystemExit→5, and flag requires JSON, but not: `check` `block`→4, doctor `blocked`→4, `status`/`submit` terminal job states→4, non-strict mode still exiting `0` on warn/block (compat regression guard), or Exception-path retryable→6. Without those, the success-path and failure-path matrices can drift.

**Suggestion:** Add focused `main([... "--output", "json", "--strict-exit-codes"])` cases for block/doctor/job-failed/retryable exception, plus one non-strict assert that `check` warn still returns process exit `0` with `status=warn`.

tests/test_gemini_translate_batch_cli_contract.py
```diff
@@ -165,9 +171,83 @@ def test_json_mode_turns_system_exit_into_error_envelope(self):
...
+    def test_strict_exit_codes_requires_json_output(self):
+        stderr = io.StringIO()
+
+        with contextlib.redirect_stderr(stderr):
+            with self.assertRaises(SystemExit) as raised:
+                batch.main(["doctor", "--strict-exit-codes"])
+
+        self.assertEqual(raised.exception.code, 2)
+        self.assertIn("requires --output json", stderr.getvalue())
     def test_text_mode_preserves_human_stdout(self):
```
hu-border-collie, line 408: **[nit]** Missing blank line between `test_strict_exit_codes_requires_json_output` and `test_text_mode_preserves_human_stdout` (PEP8 E301). Harmless at runtime but inconsistent with the rest of the file.

**Suggestion:** Insert a blank line between the two methods.

cli_contract.py
hu-border-collie, line 47: **[nit]** Marker `"quota"` is a short substring and will false-positive on words like `"quotation"`. Low practical risk in this repo, but the same style of matching is what makes the whole classifier brittle.

**Suggestion:** Prefer token/boundary-aware checks or structured codes; if keeping substrings, use more specific phrases (`"quota exceeded"`, `"resource_exhausted"`, `"429"`).

hu-border-collie (commented):
## Summary

PR #278 adds opt-in `--strict-exit-codes` (JSON-only) plus message-based `classify_error` / `strict_exit_code` mapping for Agent-friendly process exits. The success-path mapping for `check` (`safe`/`warn`/`block`) and doctor/job terminal states looks correct and is covered by unit tests; default process exit codes remain `0`/`1`/`2` when the flag is off. The dominant risk is that error classification is substring matching on human-readable text and does not align with the real API/error strings already recognized elsewhere in the codebase, so the documented `REMOTE_RETRYABLE` / exit `6` path and several apply-preflight exit-`5` cases will often not fire. JSON-mode `error.code` also changes for all machine-output failures even without `--strict-exit-codes`.

## Issue counts by severity

- bugs: 2
- suggestions: 3
- nits: 2

PR #291, opened 2026-08-01
Implement Ren'Py P2 adapter writeback plan

hu-border-collie:
## Summary

- Implement Ren'Py P2 occurrence relocation using identity v2 first, then unique source/context evidence fallback.
- Add stable adapter validation reason codes for tags, fields, percent tokens, and placeholders.
- Add declarative text-span writeback plans with common source fingerprint, path containment, span, hash, overlap, and plan-digest checks.
- Route sync preview, Batch, and revision flows through the common adapter-plan gate while retaining existing check -> apply, source re-read, transaction, and atomic-write safety.
- Persist and revalidate the writeback plan for sync preview apply.
- Add P2 regression coverage and update the adapter contract documentation.

## Why

#265 P2 requires engine-specific relocation and format validation without duplicating writeback safety in each workflow. This keeps Ren'Py rendering semantics in the adapter while keeping project, manifest, path, stale-source, transaction, and writeback safety in the common layer.

## Validation

- Focused adapter, sync preview, identity, Batch, and runtime regression suites passed.
- `python -B tests/run_cli_tests.py -q` — 854 tests passed, 1 skipped.
- `python -B tests/run_gui_tests.py -q` — 900 tests passed.
- `python scripts/run_quality_gates.py all` — all blocking gates passed.

Refs #265

P3/P4 and later epic phases remain outside this PR.

## Summary by CodeRabbit

* **新功能**
  * 支持同步翻译预览、普通批量翻译及修订预览/应用流程。
  * 新增文本定位、占位符与格式校验，以及声明式写回计划。
  * 预览结果记录失败详情，并区分安全与部分完成状态。

* **安全性改进**
  * 应用前重新验证源文件，检测过期、冲突、重叠或不安全操作并阻止写入。
  * 支持源文件变化后的可靠定位，同时保留 UTF-8 BOM 等文件特性。

* **文档**
  * 更新引擎适配器能力、工作流及安全边界说明。

### DIFF
--- a/engine_adapters/writeback.py
+++ b/engine_adapters/writeback.py
@@ -10,7 +10,7 @@ from __future__ import annotations
 
 import hashlib
 import ntpath
-from typing import Mapping, Sequence
+from typing import Mapping, NoReturn, Sequence
 
 from .contracts import (
     WRITEBACK_PLAN_SCHEMA_VERSION,
@@ -73,7 +73,7 @@ def _operation_payload(operation: WritebackOperation) -> dict:
     return payload
 
 
-def _fail(reason_code: str, message: str) -> None:
+def _fail(reason_code: str, message: str) -> NoReturn:
     raise WritebackPlanError(reason_code, message)
 
 
--- a/engine_adapters/writeback.py
+++ b/engine_adapters/writeback.py
@@ -1,0 +1,259 @@
+"""Common validation and rendering for declarative adapter writeback plans.
+
+Adapters describe safe text-span operations.  This module re-checks the plan
+against the live source snapshot and renders changed lines in memory.  It has
+no filesystem write authority; workflow code remains responsible for path
+resolution, transaction journaling, and atomic writes.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import ntpath
+from typing import Mapping, Sequence
+
+from .contracts import (
+    WRITEBACK_PLAN_SCHEMA_VERSION,
+    SourceDocument,
+    WritebackOperation,
+    WritebackPlan,
+)
+from .coverage import digest_json
+
+
+class WritebackPlanError(ValueError):
+    """Raised when a declarative writeback plan is not safe to consume."""
+
+    def __init__(self, reason_code: str, message: str):
+        super().__init__(message)
+        self.reason_code = reason_code
+
+
+def source_snapshot_fingerprint(source_documents: Sequence[SourceDocument]) -> str:
+    """Return the common source snapshot digest used by adapter plans."""
+
+    return digest_json(
+        [
+            document.manifest_entry()
+            for document in sorted(source_documents, key=lambda item: item.file_rel_path)
+        ]
+    )
+
+
+def _sha256_text(value: str) -> str:
+    return hashlib.sha256(value.encode("utf-8")).hexdigest()
+
+
+def _normal_relative_path(value: str) -> str:
+    raw = str(value or "")
+    if not raw or ntpath.isabs(raw) or ntpath.splitdrive(raw)[0]:
+        raise WritebackPlanError(
+            "common.writeback.path_escape",
+            f"Writeback target must be a relative path: {value!r}",
+        )
+    normalized = raw.replace("\\", "/")
+    parts = normalized.split("/")
+    if any(not part or part == "." or part == ".." for part in parts):
+        raise WritebackPlanError(
+            "common.writeback.path_escape",
+            f"Writeback target contains an unsafe path component: {value!r}",
+        )
+    return "/".join(parts)
+
+
+def _plan_payload(plan: WritebackPlan) -> dict:
+    payload = plan.to_dict()
+    payload.pop("plan_digest", None)
+    return payload
+
+
+def _operation_payload(operation: WritebackOperation) -> dict:
+    payload = operation.to_dict()
+    payload.pop("operation_id", None)
+    return payload
+
+
+def _fail(reason_code: str, message: str) -> None:
+    raise WritebackPlanError(reason_code, message)
+
+
+def _validate_operation(
+    operation: WritebackOperation,
+    documents: Mapping[str, SourceDocument],
+    spans: list[tuple[str, int, int, int]],
+) -> tuple[str, SourceDocument, list[str]]:
+    if operation.kind != "text_span_replace":
+        _fail(
+            "common.writeback.operation_unsupported",
+            f"Unsupported writeback operation kind: {operation.kind!r}",
+        )
+    if operation.target_root != "localization_catalog":
+        _fail(
+            "common.writeback.target_root_unsupported",
+            f"Unsupported writeback target root: {operation.target_root!r}",
+        )
+    if any(
+        not str(value or "")
+        for value in (
+            operation.occurrence_id,
+            operation.expected_file_sha256,
+            operation.expected_fragment_sha256,
+            operation.expected_text_digest,
+            operation.validation_digest,
+        )
+    ):
+        _fail(
+            "common.writeback.plan_invalid",
+            f"Writeback operation is missing required integrity fields: {operation.operation_id!r}",
+        )
+    if not operation.operation_id.startswith("op1:"):
+        _fail(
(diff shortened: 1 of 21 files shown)


gemini_translate_batch.py
```diff
@@ -12541,3 +12942,5 @@ def main(argv=None):
 
 if __name__ == '__main__':
     raise SystemExit(main())
+    Occurrence,
+    OpaqueLocator,
```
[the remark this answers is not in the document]
hu-border-collie: Fixed in 249cea3. `Occurrence` and `OpaqueLocator` were already present in the contracts import block, so the stray unreachable expressions after `raise SystemExit(main())` were removed. Explicit Ruff B018 now passes.

translator_runtime.py
```diff
@@ -5102,7 +5262,11 @@ def run_translation(*, prepare=False):
         print(f"Sync preview report: {report_path}")
         print(f"Preview files: {int(summary.get('files_changed') or 0)}")
         print(f"Preview translations: {int(summary.get('translated_items') or 0)}")
-        print("Preview status: safe")
+        if preview_failures:
+            print(f"Preview failures: {len(preview_failures)}")
+            print("Preview status: partial")
```
hu-border-collie, line 5267: **[bug]** Sync preview can now print Preview status: partial when per-file adapter plan build fails for some files while others succeed (exit code still 0, manifest still written). GUI still hard-requires the literal string Preview status: safe in gui_qt/sync_translation_report.py before treating a preview as successful, and existing GUI tests only cover safe. A valid partial preview will therefore be summarized as failed in the GUI even though CLI produced an applyable manifest and set_preview_ready may still arm apply on exit code 0 — inconsistent user-facing behavior. Agents.md requires user-facing CLI/GUI status contracts to stay in sync.

**Suggestion:** Update GUI summarizer (and its tests) to accept Preview status: partial as a warning/partial-success state (or keep printing a status line the GUI already understands and surface failures only via Preview failures: / summary fields). Ensure apply affordances and workflow status agree on partial previews.

gemini_translate_batch.py
```diff
@@ -7533,29 +7970,29 @@ def apply_results(target=None, force=False):
...
+    if adapter_plan is not None and adapter_snapshot is not None:
+        rendered_by_file = render_writeback_plan(
+            adapter_plan,
+            adapter_snapshot.project.source_documents,
         )
+        for file_key in revalidated_replacements_by_file:
+            if not revalidated_replacements_by_file[file_key]:
+                continue
+            rendered_lines = rendered_by_file.get(file_key)
+            if rendered_lines is not None:
```
hu-border-collie: **[bug]** After a successful adapter plan, apply_results only appends a write when rendered_by_file.get(file_key) is not None. If a file still has revalidated replacements but is missing from the rendered map (path-key mismatch between manifest file_key and plan target_rel_path, or any future plan/render divergence), the loop silently skips the write and later still update_progress that file as applied. That is fail-open relative to the writeback safety model: progress can claim success without mutating the script. The same if rendered_lines is not None pattern exists in apply_revisions.

**Suggestion:** If any revalidated replacement file is absent from rendered_by_file, treat it as adapter_writeback_block / refuse apply before any write or progress update. Prefer normalizing both sides to the same relative path helper so keys cannot diverge.

gemini_translate_batch.py
```diff
@@ -6330,6 +6370,266 @@ def validate_result_replacements(manifest, replacements_by_file, summary):
...
+        ]
+        if len(by_id) == 1:
+            return by_id[0]
+
+    positioned = [
+        occurrence
+        for occurrence in occurrences
+        if occurrence.unit.file_rel_path == file_key
+        and occurrence.unit.line == line
+        and occurrence.unit.start == start
```
hu-border-collie: **[suggestion]** _find_adapter_occurrence returns a unique identity match without verifying that the live occurrence span/text agrees with the replacement's already source-validated (line, start, end, expected_text). Plan write spans then come from the occurrence, while validate_replacements_for_lines previously authenticated the replacement coordinates. Under normal relocate+extract agreement this matches, but the two coordinates are not bound together at the plan gate.

**Suggestion:** After an id hit, require occurrence.unit.line/start/end (and expected text in {text, source_text, current_translation}) to match the replacement; otherwise fall through to the positioned matcher or fail closed with common.locator.unresolved.

engine_adapters/writeback.py
```diff
@@ -0,0 +1,259 @@
...
+            "common.writeback.span_invalid",
+            f"Writeback span is outside the live line: {rel_path}:{operation.line}",
+        )
+    if "\n" in operation.replacement_fragment or "\r" in operation.replacement_fragment:
+        _fail(
+            "common.writeback.replacement_invalid",
+            f"Writeback replacement must stay on one line: {rel_path}:{operation.line}",
+        )
+    raw_fragment = line[operation.start_col : operation.end_col]
+    if _sha256_text(raw_fragment) != operation.expected_fragment_sha256:
```
hu-border-collie, line 159: **[suggestion]** Common plan validation re-checks expected_fragment_sha256 against the live raw span bytes but never re-checks expected_text_digest against the decoded string literal at that span. Fragment hash is usually sufficient, yet the field is part of the operation integrity payload and could detect decode/normalization mismatches if fragment hashing and literal evaluation ever disagree (or if a plan is hand-edited with a consistent fragment hash but inconsistent text digest before operation_id re-signing).

**Suggestion:** After the fragment hash check, decode the live span (or reuse token literal evaluation) and require _sha256_text(decoded) == expected_text_digest, or document that expected_text_digest is payload-binding only and not a live check.

hu-border-collie (commented):
## Summary

This PR delivers a solid P2 adapter writeback design: Ren'Py relocation/validation stay in the adapter, declarative text_span_replace plans are revalidated by a common gate (source fingerprint, span/hash/overlap, path normalization, plan digest), and Batch/sync/revision still route actual writes through existing check->apply / preview fingerprint / atomic-write paths. The dominant residual risks are a GUI contract break for the new sync Preview status: partial outcome, a fail-open apply path if rendered plan keys ever diverge from replacement file keys, and occurrence resolution that trusts identity match without re-checking the replacement span that was source-validated.

## Issue counts by severity

- bugs: 2
- suggestions: 4
- nits: 1

## Issues outside the diff

These findings reference lines that are not present in the diff and could not be posted as inline comments:

- **[suggestion]** engine_adapters/renpy.py:600 -- Content-evidence relocation accepts a unique highest score when identity v2 misses. Tests intentionally flipped a former v2_relocation_missing hard block into a successful unique-content match. That is fail-open relative to strict identity: a stale id with a still-unique source string will write wherever that string uniquely appears (score prefers same file/speaker/block, but only uniqueness of the top score is required). Residual mis-apply risk if a string is unique for the wrong reason (e.g. duplicates removed elsewhere).
  - **Suggestion:** Optionally require a minimum absolute score and/or same-file constraint for content_evidence; keep ambiguous ties unresolved (already done). Add a regression that two same-score candidates still block writeback end-to-end through check/apply, not only unit relocation.
- **[suggestion]** gemini_translate_batch.py:1419 -- _validate_adapter_writeback_plan catches broad (Exception, SystemExit) and converts them into adapter_writeback_block. That correctly refuses writeback, but collapses distinct hard failures (path/project SystemExit, unexpected bugs) into a soft plan block reason, which can make diagnosis and safety reporting less precise.
  - **Suggestion:** Catch WritebackPlanError and expected ValueErrors explicitly; let SystemExit and unexpected exceptions propagate (or map known project-mismatch exits to common.writeback.project_mismatch without swallowing unrelated exits).
- **[nit]** tests/test_engine_adapter_p2.py:2556 -- P2 unit tests cover path escape, source snapshot mismatch, span/source mismatch, and non-pass validation well, but there is no direct test that a tampered plan_digest / operation_id is rejected by validate_writeback_plan, nor that Batch apply --force still cannot write when the adapter plan is block/stale. Those are claimed common-gate invariants.
  - **Suggestion:** Add focused tests for digest tampering and for apply/force refusal when adapter_writeback_block is present after revalidation.

PR #371, opened 2026-08-16
fix(gui): defer workspace dialog SDK stop off the UI thread (#297 P1)

hu-border-collie:
Parent: #297（本 PR 完成 P1 的最后一个缺口）

## 背景

P0（PR #334）落地 shutdown coordinator 时已超额覆盖了 P1 的大部分语义：

- doctor 取消后经 `_retired_doctor_workers` 保留对象到真实 `finished`；
- registry `hideEvent()` 已改为非阻塞 `request_stop_active_workers()`，worker `completed` 必发、对象在真实 `finished` 后 `deleteLater`；
- LiteLLM warmup 有应用级 retired 集合；font / context / optional feature 均为「父子所有权 + deleteLater + shutdown 守卫」模式；
- coordinator 的 `stalled` 只作提示、轮询继续、窗口保持打开，bounded 超时不丢引用伪装完成（含 `user_copy` 的 stalled / 云任务区别文案）。

本 PR 补上 P1 剩余的最后一个同步等待点：首次运行「创建 / 接入工作区」向导在 SDK 下载期间，取消（reject）或关闭（closeEvent）会调用 `QThread.wait(5000)`，GUI 线程最长冻结 5 秒。

## 改动

- `gui_qt/workspace_setup_dialog.py`
  - 删除阻塞的 `_stop_sdk_worker()`；新增 `_request_sdk_worker_stop(origin)`：只发起协作取消（`request_cancel` + `requestInterruption`），拒绝/关闭延迟到 worker 真实 `finished` 信号再从事件循环完成。
  - `_sdk_worker_active()` 改为引用语义（引用存在即未终态），修掉了 `completed` 与真实 `finished` 之间窗口期内对话框可被销毁、线程仍在收尾的隐患。
  - worker 创建时接线 `finished → deleteLater` 与终态处理；`_on_sdk_completed` 在延迟停止期间让位（不再争抢状态与按钮）。
- `gui_qt/user_copy.py`：新增 `WORKSPACE_SETUP_STOP_COPY`（取消中 / 等待关闭 / 仍在结束 / 已取消）。
- `tests/test_gui_workspace_setup_dialog.py`：新增确定性假 worker（无真实线程）与 5 个用例：延迟拒绝不阻塞、真实 `finished` 后结算、pending 期间 `completed` 被抑制、引用保留到真实 `finished`、close 延迟并自动完成。
- `docs/gui_workbench.md`：项目列表分区补充向导取消/关闭的非阻塞行为说明。

## 验证

[...]

PR #373, opened 2026-08-17
fix(gui): drop stale async results by operation identity (#297 P2)

hu-border-collie:
Parent: #297（本 PR 完成 P2：operation identity / stale-result）

仅 GUI：不改变翻译 / manifest / writeback 语义。Doctor、模型目录、字体、registry 仍沿用 P1 的 worker ownership，未改挂这条 digest 合同。

## 背景

P0（#334）和 P1（#371）已把关闭路径和 retired worker 从 GUI 线程同步等待里拆出来。剩下的 P2 缺口是：异步结果没有统一身份，迟到的回调仍可能覆盖新项目 / 新配置 / 新连接参数。

Issue 点名的两处：

- 上下文扫描只按项目路径挡旧结果，未绑定配置 digest
- LiteLLM 连接测试结果未绑定 provider / model / 自定义端点

## 改动

- `gui_qt/operation_identity.py`：共享 `canonical_digest()` / `is_current_identity()`，以及上下文配置 digest 与 LiteLLM 连接 identity（provider / model / 自定义端点，不含密钥）
- `gui_qt/context_library_worker.py`：扫描结果携带 `config_digest`
- `gui_qt/app.py`
  - 上下文库完成回调同时比对项目路径和配置 digest；过期结果不写缓存、不改页面，并在项目/配置已变时再排一次后台刷新
  - LiteLLM 连接测试启动时写入 identity；迟到的旧 worker 只做清理；参数已变时丢弃结果
  - 网络进度也按当前 worker 过滤
- `gui_qt/user_copy.py`：新增 `LITELLM_CONNECTION_TEST_COPY["stale_result"]`
- `docs/gui_workbench.md`、`CHANGELOG.md`：同步用户可见行为

## 验证

- `python -m unittest tests.test_gui_operation_identity tests.test_gui_context_library_worker tests.test_gui_litellm_worker tests.test_gui_litellm_settings_page tests.test_gui_task_pages -q` → 通过
- `python -m unittest tests.test_gui_cli_runner tests.test_gui_doctor_worker tests.test_gui_games_registry_worker tests.test_gui_context_library_worker tests.test_gui_litellm_worker tests.test_gui_lifecycle -q` → 77 通过
- `python -B tests/run_gui_tests.py -q` → 1114 个用例里 1113 通过；失败的 `test_workspace_action_bar_uses_immediate_save_context` 在当前 `main` 上单独跑也同样失败，不是本 PR 引入
- `git diff --check` → 干净


gui_qt/operation_identity.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    result = str(result_identity or "")
+    if not result:
+        return True
+    return result == str(current_identity or "")
+
+
+def context_library_config_digest(config: object) -> str:
+    """Digest the translator-config snapshot used for a context-library scan."""
+    return canonical_digest(config or {})
```
hu-border-collie: **[bug]** `context_library_config_digest` hashes only the in-memory `translator_config` snapshot, but the scan's user-visible switches come from `project_context_settings.json` via `read_batch_context_flags(config, game_root=base_dir)`. `_on_save_config` deliberately does not write RAG / source-index / project-analysis enablement into `translator_config.json` (it restores `batch.project_analysis.enabled` / `inject_published_brief` to the pre-save global values and persists the checkboxes only in the project file). After a flags-only save the translator-config digest is unchanged, so `_on_context_library_status_ready` treats the in-flight result as current, writes `_context_library_status_cache` / `_context_library_flags_cache` from the old `context_flags`, and skips the new same-path pending refresh. The post-save `_refresh_context_library_panel_async()` can also TTL-return without even setting `pending`, because the old cache still matches the config digest. This is the named #297 P2 gap (scan keyed only by project path) and contradicts CHANGELOG / `docs/gui_workbench.md` ("保存设置后，旧扫描不会用过期开关覆盖当前页"). Hashing the whole translator_config is also too wide: a theme/model-only save will false-stale a still-valid scan.

**Suggestion:** Digest the inputs the worker actually used, not the raw global file. At collect time hash `{effective_flags, config_or_relevant_subset}` where `effective_flags` is the `read_batch_context_flags(...)` result (or the project file payload). At apply / cache-invalidation time compute the same digest from the current snapshot **plus** `_context_library_flags_cache` / a fresh flag read after save. Do not hash unrelated `translator_config` keys (theme, sync models). Add a GUI test: page scan in flight → toggle `rag_enabled` and save → stale result must not replace flags/label, and a force refresh must be scheduled.

gui_qt/operation_identity.py
```diff
@@ -0,0 +1,87 @@
...
+    provider: str,
+    model: str,
+    custom_providers: Mapping[str, object] | None = None,
+) -> str:
+    """Digest the provider/model/custom-provider state a connection test used."""
+    return canonical_digest(
+        {
+            "provider": str(provider or "").strip().lower(),
+            "model": str(model or "").strip(),
+            "custom_providers": custom_providers or {},
```
hu-border-collie: **[suggestion]** `litellm_connection_identity` folds the entire `custom_providers` map into the digest. `_on_test_litellm_connection` only needs the **selected** provider's rewrite/credential fields (`_custom_litellm_providers.get(provider)` → `base_url` / `models_url` / `requires_key` / `api_key_env`). This is not a false-apply bug: a valid test for `openai/gpt-x` is dropped if the user adds or edits an unrelated custom row on the same LiteLLM page while the request is in flight. That is a realistic same-page action, not a theoretical race. Secrets are not leaked (`CustomLiteLLMProvider` has no key material; `api_key_env` is a name). Acceptable as a conservative stale-drop, but wider than the named contract (provider / model / endpoint of **this** test).

**Suggestion:** Hash `provider`, `model`, and only the selected id's endpoint fields (or `{}` for a built-in). Ignore `label` and every other registry row. Add a unit test that mutating an unused custom provider keeps the identity, while changing the selected `base_url` does not.

gui_qt/operation_identity.py
```diff
@@ -0,0 +1,87 @@
+"""Stable operation identities for asynchronous GUI results.
+
+Issue #297 P2 requires every async task to carry an identity (project,
```
hu-border-collie: **[suggestion]** The new module docstring narrates Issue #297 P2 and the architecture history ("every async task… stale result can only finish cleanup"). `gui_qt/user_copy.py` does the same (`#297 P2`; "the result itself is dropped"). Comments here should say why a digest is compared, not retell the epic.

**Suggestion:** Keep a one-line why (e.g. worker and window must compute the same opaque id independently). Drop issue numbers and the change narrative from `operation_identity.py` and the `stale_result` copy comment.

hu-border-collie (commented):
Submitted review: context-library digest misses project enablement flags (bug); LiteLLM identity is conservatively too wide; trim epic-narrating comments.

hu-border-collie:
Review follow-up is in.

- Context-library digest hashes effective enablement flags, not the whole `translator_config` (`10d7f52`).
- LiteLLM identity hashes only the selected provider endpoint (`10d7f52`).
- Epic-narrating comments trimmed (`10d7f52`).
- Digest no longer reads `_context_library_flags_cache`, so a stale applied-flags snapshot cannot mismatch a new worker result and reschedule forever (`a234cef`).

P2 still only covers the two named gaps (context scan + LiteLLM connection test). Fonts / optional feature / generation stay on later #297 stages.

The three inline review threads are resolved.

PR #377, opened 2026-08-20
feat(model): add #345 P3 routing preflight and diagnostics

hu-border-collie:
## Summary

- validate only the active TaskRoute stages before prepare, package creation, provider calls, submit, and probe side effects
- expose credential-safe sync and Gemini Batch routing diagnostics through doctor with stable machine error contracts
- route the GUI LiteLLM connection test through the shared ModelProfile resolver and backend factory
- document the new error codes and doctor state, and update legacy LiteLLM test fixtures to use provider-qualified models

Refs #345

## Behavior changes

- unsupported profile/strategy combinations now fail before a task starts
- unrelated optional stages do not block the current task
- retries remain on the frozen profile; no provider or model fallback is introduced
- diagnostic API keys remain ephemeral and are never serialized into profiles, manifests, logs, or doctor output

## Validation

- python -B tests/run_cli_tests.py -q (1381 passed, 1 skipped)
- python -B tests/run_gui_tests.py -q (1123 passed)
- python scripts/run_quality_gates.py all
- 41 tracked Markdown files checked for valid local links
- git diff --check


gemini_translate_batch.py
```diff
@@ -6722,6 +6731,28 @@ def submit_manifest(
     if model_override:
         manifest['batch_model'] = model_override.strip()
 
+    submit_stage = model_profile.stage_for_manifest_mode(manifest_mode(manifest))
+    if model_override:
+        submit_plan = model_profile.resolve_routing_plan_from_runtime(
```
hu-border-collie: **[bug]** `submit --model` rebuilds a full `ModelRoutingPlan` from *current* runtime globals and then applies `stage_overrides={submit_stage: model_override}`. For translation/keyword/revision, `resolve_routing_plan` builds that override with `_sync_profile(..., sync_backend=SYNC_BACKEND)`, not `_batch_profile`. When `sync.backend` is `litellm` (the mixed LiteLLM-sync + Gemini-batch setup this architecture exists for), a valid unprefixed Gemini id such as `gemini-2.5-flash` becomes a LiteLLM profile with an empty provider prefix. `require_valid_routing_plan` then refuses with `MODEL_ROUTE_CAPABILITY_MISSING` / `MODEL_PROFILE_INVALID` *before* upload, so a submit override that previously only rewrote `manifest['batch_model']` now fails. Even when `SYNC_BACKEND` is `gemini` and validation passes, `attach_model_routing` overwrites the package-time snapshot with current `SYNC_MODEL`, `PROJECT_ANALYSIS_MODEL`, `FINAL_REVIEW_MODEL`, and `config_origins`, which breaks the frozen-run contract for every non-submit stage in that manifest.

**Suggestion:** Do not reconstruct the plan from live runtime settings. Start from `resolve_manifest_routing_plan(manifest, execution=GEMINI_BATCH)`, replace only the submit-stage profile with `_batch_profile(model_override)` (or set `batch_model` without a sync-adapter stage override), validate `{submit_stage}`, and attach that patched snapshot. Add a regression test: LiteLLM `SYNC_BACKEND` + `submit --model gemini-2.5-flash` must stay `gemini_adapter` and must not rewrite unrelated stage profiles.

gui_qt/litellm_worker.py
```diff
@@ -483,10 +483,38 @@ def _cancel_active_async_task(self) -> None:
...
+            custom_providers=self._custom_providers,
+            # Connection testing is itself the credential probe. A typed key
+            # is ephemeral, while an empty value may still resolve from the
+            # normal keyring/environment path inside the backend.
+            keyring_has_credential=None,
+        )
+        if issues:
+            raise LiteLLMBackendError(
+                issues[0].message,
+                category="unsupported_capability",
```
hu-border-collie: **[bug]** Routing preflight failures in the connection test are raised as `LiteLLMBackendError(..., category="unsupported_capability")` without `internal=True`. Two layers then discard the useful `RoutingValidationIssue`: `LiteLLMBackendError` replaces any non-whitelisted message with "LiteLLM provider does not support this request.", and `_connection_error_message` maps the category to `LITELLM_CONNECTION_TEST_COPY["errors"]["unsupported_capability"]` ("当前 Provider 不支持连接测试所需的 JSON 能力。"). A typed model missing the `provider/model` prefix, an unknown adapter, or a capability-override error therefore looks like a JSON-schema incompatibility. `validate_routing_plan` does not even check structured-output support, so this category is the wrong contract for almost every issue this branch can emit. Existing GUI tests only retarget the `build_sync_backend` patch; they never exercise a failing preflight.

**Suggestion:** Map `MODEL_PROFILE_INVALID` / `MODEL_ROUTE_CAPABILITY_MISSING` to distinct GUI copy (or at least `provider_error`) and pass `internal=True` if the issue message should survive. Add a worker test that constructs `LiteLLMConnectionTestWorker("gpt-4o-mini")` (no provider prefix) and asserts the completed text mentions the prefix rule rather than JSON capability.

gemini_translate_batch.py
```diff
@@ -14536,6 +14609,49 @@ def collect_doctor_project_assets_warnings(project_assets):
...
+            payload['execution_strategy'] = strategy.value
+            key = (
+                payload['code'], payload['stage'], payload['profile_id'],
+                tuple(payload['missing_capabilities']), strategy.value,
+            )
+            if key not in seen:
+                seen.add(key)
+                issues.append(payload)
+    return {
+        'status': 'blocked' if issues else 'ok',
```
hu-border-collie: **[suggestion]** `collect_doctor_model_routing_status` sets `status` to `'blocked'` whenever *any* stage of *either* `sync` or `gemini_batch` plan has an issue. Doctor still only appends warnings and does not change the command exit/workflow gate, which matches the new docs, but the machine field name collides with doctor/layout `blocked` and CLI envelope `status=blocked` (exit 4). An agent that short-circuits on any `"status": "blocked"` will refuse a sync translation because unused final-review/batch config is invalid — the exact false block this PR claims to avoid at task start.

**Suggestion:** Use `'ok' | 'attention'` (or echo the primary issue code) for `model_routing.status`, and document that it never changes doctor exit. Keep `'blocked'` for task-start `routing_validation_error` only.

gui_qt/user_copy.py
```diff
@@ -737,6 +737,8 @@ def translate_doctor_warning(warning: str) -> str:
         return "自定义模板命令无法解析，请检查配置。"
     if text.startswith("RAG store contains legacy ID format keys."):
         return "记忆库含有旧版键格式，下次写回时会自动迁移。"
+    if text.startswith("Model routing preflight ["):
+        return "模型路由启动前检查失败；请检查 Provider、模型、执行方式和凭据引用。"
```
hu-border-collie, line 763: **[suggestion]** GUI doctor translation collapses every `Model routing preflight [CODE] (strategy/stage): message` line into one generic sentence. CLI/JSON still expose the stable code, strategy, and stage, but the GUI is the user-facing surface and now hides whether the problem is a bare LiteLLM model, a Gemini-batch-on-LiteLLM strategy, a missing custom-provider env/keyring ref, or an unused final-review stage. Combined with the doctor `status: blocked` naming, a non-blocking unused-stage warning will read as “routing failed, do not start”.

**Suggestion:** Keep the prefix mapping, but include code, strategy, and stage in the Chinese text (and optionally a short per-code hint). A test that only `assertIn("模型路由启动前检查失败")` will not catch this loss of detail.

model_profile.py
```diff
@@ -1151,7 +1177,8 @@ def validate_routing_plan(
         elif ref.kind == CREDENTIAL_KIND_KEYRING:
             env_missing = not (ref.env_name and ref.env_name in env_lookup)
             keyring_missing = (
-                keyring_has_credential is not None
+                (custom is not None or bool(ref.env_name))
```
hu-border-collie, line 1221: **[suggestion]** Keyring emptiness is now probed only when `custom is not None or bool(ref.env_name)`. Built-in LiteLLM profiles are `CredentialRef(KEYRING, provider)` with an empty `env_name`, so `openai/gpt-4o-mini` never emits `MODEL_PROFILE_CREDENTIAL_REF_MISSING` even though production now always passes `_runtime_keyring_has_credential`. That is probably required to avoid false positives when LiteLLM reads `OPENAI_API_KEY` from the environment, but the validator docstring still says a supplied probe that reports empty is enough, and there is no comment on the new gate. Doctor and task preflight therefore will not catch a missing built-in key before the first request.

**Suggestion:** Document the gate (custom/env_name refs are machine-decidable; built-in providers also have native env vars this module does not model). If built-in missing keys should be in-scope, consult LiteLLM's env mapping or treat constructor/env fallback as a third source instead of skipping the probe.

hu-border-collie (commented):
Stage-scoped routing preflight, doctor secret hygiene, and the GUI LiteLLM test going through the shared resolver look correct. Remaining correctness issues: `submit --model` rebuilds the Gemini Batch plan through the sync adapter, and GUI connection-test preflight failures are mapped to `unsupported_capability`.

PR #380, opened 2026-08-22
feat(plan): translation plan core + golden fixture (#346 P1 + P0 tails)

hu-border-collie:
Refs #346（P0 尾巴 + P1 纯核心）。按 [实施计划](https://github.com/hu-border-collie/renpy-translation-lab/blob/docs/issue-346-implementation-plan/docs/plans/issue-346-implementation-plan.md) §4 与 [D1–D7 定稿评论](https://github.com/hu-border-collie/renpy-translation-lab/issues/346#issuecomment-5379565547) 实施，不含 P2/P3 执行器接线。生产委托改动中，chunking、hash key、局部上下文保持行为不变；唯一行为变化：sync 词法术语（normalize_map 命中）渲染顺序改为按键排序的稳定顺序（#346 确定性设计，内容不变、仅 glossary 行序，已在 CHANGELOG「变更」区披露）。

## P0 两个尾巴（随本 PR 交付）

1. **计划文档状态行**：`docs/plans/issue-346-implementation-plan.md` 从「草案（未开工）」改为「已定稿决策」，§3.1 决策表头部链接到 issue #346 的 D1–D7 定稿评论。
2. **golden fixture 骨架**：`tests/fixtures/translation_plan_minimal/`——手写最小 `.rpy` 源（两个 label 形成 block 边界素材、说话人、`[Gil_name!t]` 插值、术语命中）、固定配置与 ModelProfile、词法术语/macro/检索/分析层输入，以及冻结的 `expected/plan.sync.json` + `plan.gemini_batch.json`。

## P1 核心内容

- **`translation_plan.py`（新增，纯模块）**：不 import SDK/网络/执行器（子进程纯度测试断言）。
  - schema v1：`TranslationPlan` / `PlanChunk` / `TranslationRequest` / `SourceIdentity` / `ContextAssembly`，`to_dict`/`from_dict` 可序列化。
  - 稳定 ID：`chunk_id` 保留 batch 现格式 `sha1(path)[:10]-NNNNN`；`request_id = sha256(canonical(plan_id, chunk_id, expected_ids))[:16]`。
  - 三级指纹：`plan_id`（内容身份，排除 run_id 与检索内容）、`prompt_fingerprint`（模型语义合同，跨策略可比）、`request_fingerprint`（含 generation/transport 审计全量）；全部由 `canonical_json` 内容派生，无时间戳/随机数。
  - `redact_sensitive()`：api_key/authorization/token/secret/password 形状的键值整体替换；`credential_ref` 是引用、放行。
- **`translation_core.py`（增量）**：D4 共享 chunking `iter_translation_chunk_ranges`（60/18000 默认）+ `file_hash_key`；D3 canonical prompt——system instruction 以 batch 形态为基合并 Sync 独有规则（人名保留英文、禁 markdown/Pinyin），user prompt 分段组合（词法术语块 + reference 块 + CONTEXT/TARGET/compact JSON），D5 默认常量冻结。
- **六层上下文组装**：required/local/project/retrieval/analysis 五内容层按 rank 排序、重复文本确定性丢弃并记录原因、检索层预算（history 220 + story 1200）确定性截断；预算层为组装级记账。D1 block 边界窗口与 D2 词法术语命中从 `translator_runtime` 上移为共享实现。
- **生产委托（行为不变）**：`gemini_translate_batch.hash_key` / `iter_translation_chunk_ranges` / `task_text_char_count`、`translator_runtime.build_sync_local_context` / `retrieve_sync_glossary_hits` 改为委托共享实现；旧 prompt builder 保留为兼容 shim。

## 黄金等价的前身

fixture 上 sync 与 gemini_batch 构建：`plan_id` 不同（策略在身份内）、但**每个 chunk 的 `prompt_fingerprint` 完全一致**、`request_fingerprint` 因 transport 差异不同——正是 P4 黄金等价测试要收口的不变量，P1 已在纯核心层锁定。

## 验证

- `python -m unittest tests.test_translation_plan tests.test_translation_core -q`：65 通过
- 既有回归：`tests.test_translator_runtime`（158）、`tests.test_batch_golden_corpus` + `test_batch_rag` + `test_batch_repair`（124）通过，证明委托零行为变化
- `python -B tests/run_cli_tests.py -q`：1428 通过（1 skip 为既有 SDK 可选跳过）
- `python scripts/run_quality_gates.py all`：全绿；`git diff --check` 干净
- fixture 跨平台稳定：`_file_digest` 做换行归一化，不受 autocrlf 影响

## 后续

- P2：`build_chunks` 改为消费 plan builder（含 manifest v2 `translation_plan` 块与 requests.jsonl 兼容、D2 解除 batch `RAG_ENABLED` 门控）。
- P3：Sync 主循环内联 batching 换 plan chunks + `SyncGenerationRequest.system_instruction`。
- P1 冻结的 provider 接缝（`retrieval_blocks_provider` / `analysis_blocks_provider` / `render_reference_blocks`）即 #341 的接入点。


translation_plan.py
```diff
@@ -0,0 +1,975 @@
...
+        chunk_input.analysis_blocks_text = analysis_text
+
+        assembly = assemble_context_layers(chunk_input, context_policy)
+        system_instruction = translation_core.build_canonical_translation_system_instruction(
+            preserve_terms,
+            macro_setting=str(macro_setting or ''),
+        )
+        reference_blocks_text = ''
+        if retrieval_text or analysis_text:
+            reference_blocks_text = retrieval_text + analysis_text
```
hu-border-collie: **[bug]** Context assembly is not the semantic contract the model sees. `_retrieval_layer` truncates to `history_char_limit + story_char_limit` and `assemble_context_layers` drops duplicate layer texts into `dropped`, but `build_translation_plan` then builds `user_prompt` from the raw `retrieval_text + analysis_text` (and untrimmed lexical/local/required text). `prompt_fingerprint` hashes both the untrimmed prompt and the assembly diagnostics, so a retrieval blob over 1420 chars is reported as `truncated=true` / `char_used=1420` while the actual prompt still contains the full text. The fixture is too small to catch this, and `test_retrieval_layer_budget_truncates_deterministically` only asserts on `assemble_context_layers`, not on `TranslationRequest.user_prompt`. P2/P3 will send the unbudgeted prompt if they consume `user_prompt` as the source of truth. Diagnostics also unconditionally claim `rendered_by: translation_core.build_reference_blocks` even when the provider passed a raw string (as the fixture does).

**Suggestion:** Drive the canonical user prompt from the kept, budgeted layer texts (join retrieval/analysis from `assembly.layers`, not the pre-assembly strings). Add a test that a >1420-char `retrieval_blocks_provider` both sets `truncated` and shortens `user_prompt` / changes `prompt_fingerprint`. Do not freeze the current naive `text[:limit]` slice as D5 if providers are supposed to go through `render_reference_blocks` (per-section history/story limits, not a combined opaque cut).

translation_plan.py
```diff
@@ -0,0 +1,975 @@
...
+        'source_identity': identity.to_dict(),
+        'config_fingerprint': config_fingerprint,
+        'model_profile_snapshot': profile_snapshot,
+        'execution_strategy': strategy,
+        'chunk_policy': chunk_policy.to_dict(),
+        'context_policy': context_policy.to_dict(),
+        'prompt_inputs': {
+            'preserve_terms': [str(term) for term in preserve_terms or []],
+            'normalize_map': dict(normalize_map or {}),
+            'non_translatable_exact': [str(term) for term in non_translatable_exact or []],
```
hu-border-collie: **[bug]** `plan_id` and glossary hit order are not stable for the production glossary type. Identity does `[str(term) for term in non_translatable_exact or []]` (and the same for `preserve_terms`); `retrieve_lexical_glossary_hits` iterates those iterables as given. `translator_runtime.NON_TRANSLATABLE_EXACT` is a `set`. List conversion of a set depends on `PYTHONHASHSEED`, so the same terms can yield different `plan_id` and `prompt_fingerprint` across processes. The module's stated invariant is that identical inputs rebuild byte-identical plans; the golden path only passes JSON lists, so this is untested. P3 is likely to pass the runtime set straight in.

**Suggestion:** Canonicalize unordered inputs before hashing and before hit iteration (keep list/tuple insertion order; sort `set`/`frozenset`). Lock it with a test that `non_translatable_exact={'B-side','A-side'}` produces the same `plan_id` / glossary block under two hash seeds, or simply always emit a sorted unique sequence for sets.

translation_plan.py
```diff
@@ -0,0 +1,975 @@
...
+def short_fingerprint(text):
+    """16-hex content fingerprint used for plan/request/prompt identity."""
+    return sha256_hex(text)[:16]
+
+
+def _is_sensitive_key(key):
+    if str(key) in _REDACTED_KEY_ALLOWLIST:
+        return False
+    lowered = str(key).lower()
+    return any(token in lowered for token in _SENSITIVE_KEY_TOKENS)
```
hu-border-collie: **[bug]** `redact_sensitive()` is the plan snapshot's secret boundary (`model_profile_snapshot.extra_headers` is copied in), but matching is underspecified relative to the docstring ("token" markers) and real header names. `_SENSITIVE_KEY_TOKENS` uses substring match on the raw key: `'api_key' in 'x-api-key'` is false, and there is no generic `token` token (only `auth_token` / `access_token`). A profile with `extra_headers={"X-Api-Key": "real-secret"}` or `{"token": "..."}` would be hashed and serialized into the plan unredacted. `Authorization` is covered; the unit test only exercises that plus `api_key` / `password`. `credential_ref` allowlisting itself is correct and should stay.

**Suggestion:** Normalize keys (lowercase, strip `_`/`-`) before matching, and treat `token` / `api-key` / `x-api-key` as sensitive. Add a test that `extra_headers={'X-Api-Key': 'secret', 'token': 'secret'}` redacts those values while `credential_ref.name` still survives.

tests/test_translation_plan.py
```diff
@@ -0,0 +1,469 @@
...
+    def test_local_layer_records_block_bounded_diagnostics(self):
+        build = build_fixture_plan(translation_plan.STRATEGY_SYNC)
+        local = next(
+            layer
+            for request in build.requests
+            for layer in request.context_assembly['layers']
+            if layer['layer'] == translation_plan.CONTEXT_LAYER_LOCAL
+        )
+        self.assertEqual(local['diagnostics']['algorithm'], 'block_bounded_window')
+        # chapter01 chunk 1 spans chapter01_start only; the hall block bounds it.
```
hu-border-collie: **[suggestion]** The golden fixture does not lock D1. Default chunking is 60/18000, so each file is a single chunk; both frozen plans have empty `context_window_spec` (`context_before_items`/`context_after_items` 0, `block_bounded_*` false). The two labels in `dialogue.rpy` sit inside TARGET, not beside it, so they never form a block boundary for the plan. `test_local_layer_records_block_bounded_diagnostics` comments that "chapter01 chunk 1 spans chapter01_start only; the hall block bounds it" and then asserts `block_bounded_after or context_after_items == 0`, which is true of every empty window.

**Suggestion:** Either shrink the fixture `ChunkPolicy` (e.g. `max_items=4`) so chapter01 splits and the second chunk is block-bounded, or drop the misleading comment and assert the real frozen spec (`context_after_items == 0` and `block_bounded_after is False`). Keep the isolated `LocalContextWindowTests`; they already cover the algorithm.

tests/fixtures/translation_plan_minimal/inputs/file_jobs.json
```diff
@@ -0,0 +1,86 @@
+[
+  {
+    "file_rel_path": "chapter01/dialogue.rpy",
+    "file_path": "game/chapter01/dialogue.rpy",
+    "tasks": [
+      {
+        "id": "chapter01/dialogue.rpy:chapter01_start:1:2fc92d42",
+        "text": "Hey [Gil_name!t], did you finish the Dawn Chorus setlist?",
+        "line": 4,
```
hu-border-collie: **[suggestion]** README says tasks are 1:1 with `game/*.rpy`, but `line` values do not match the current sources. First dialogue task is `"line": 4` while `game/chapter01/dialogue.rpy` has that string on line 7 (0-based 6). Hall lines 11/12 in `file_jobs` are the blank line and `label chapter01_hall`, not the two say-lines (14/15). `source_identity.file_digests` hash the `.rpy` files, so P4 e2e extraction will not reproduce these unit lines or the frozen `line_numbers`.

**Suggestion:** Re-derive `line` from the checked-in `.rpy` (0-based `line` per `translation_core._coerce_line`, or documented 1-based `line_number`) and regenerate the expected plans. If `.rpy` is only narrative until P4, say that in the fixture README so the mismatch is not treated as a contract.

translation_plan.py
```diff
@@ -0,0 +1,975 @@
...
+    if value is None:
+        return ''
+    if callable(value):
+        return str(value(chunk_input) or '')
+    if isinstance(value, str):
+        return value
+    raise TypeError(f'{label} must be a string or a callable(chunk_input) -> str')
+
+
+def _unit_semantic_entry(unit):
```
hu-border-collie, line 829: **[suggestion]** `_unit_semantic_entry` (hence `plan_id`) omits `block_name`, which is the field D1 actually reads from the original task dicts (it lives on `TranslationUnit.metadata`, not a first-class field). Two `file_jobs` that differ only in `block_name` can share `plan_id` while local windows, `user_prompt`, and `prompt_fingerprint` change once a chunk is a proper subset of the file. Adapter-built ids often embed the label, but this builder accepts raw jobs and hashes only id/text/speaker/path/line.

**Suggestion:** Include `block_name` (and any other D1 inputs) in the identity payload, or document that callers must encode block identity in `unit.id`. A test that flipping `block_name` with a small `max_items` changes `plan_id` would lock it.

tests/test_translation_plan.py
```diff
@@ -0,0 +1,469 @@
...
+        self.assertEqual(
+            translation_plan.canonical_json(first.plan.to_dict()),
+            translation_plan.canonical_json(second.plan.to_dict()),
+        )
+        self.assertEqual(
+            [request.to_dict() for request in first.requests],
+            [request.to_dict() for request in second.requests],
+        )
+
+    def test_run_id_is_audit_only_and_excluded_from_fingerprints(self):
```
hu-border-collie, line 361: **[suggestion]** Tests lock `run_id` exclusion from `plan_id` / `plan_fingerprint` and strategy vs `prompt_fingerprint`, but not the other advertised identity split: retrieved content must not affect `plan_id` and must affect `prompt_fingerprint`. A regression that folds `retrieval_blocks_text` into `identity_payload` would still pass the current suite.

**Suggestion:** Add a sibling of `test_run_id_is_audit_only_and_excluded_from_fingerprints` that rebuilds with different `retrieval_blocks_provider` / `analysis_blocks_provider` strings and asserts equal `plan_id`, unequal per-chunk `prompt_fingerprint` / `plan_fingerprint`.

translation_core.py
```diff
@@ -770,6 +770,125 @@ def build_sync_translation_prompt(
     )
 
 
+# --- Canonical translation contract (issue #346, decisions D3-D6) -------------
```
hu-border-collie, line 773: **[suggestion]** The new canonical-prompt block is a design-history comment (issue numbers, which path used to own which builder, why drift should become a plan diff). It restates the change rather than a non-obvious constraint. Same pattern at `gemini_translate_batch.py:3192` ("Shared chunking algorithm (issue #346 D4); batch passes its own limits.") and the long `translation_plan.py` module docstring.

**Suggestion:** Drop the history blurb; keep a one-liner if needed, e.g. that legacy `build_translation_*` / `build_sync_translation_prompt` remain shims and P2/P3 must not call them. Constants can stay with the D4/D5 tags if those names are the frozen decision ids.

tests/test_translation_plan.py
```diff
@@ -0,0 +1,695 @@
...
+GAME_DIR = FIXTURE_DIR / 'game'
+EXPECTED_DIR = FIXTURE_DIR / 'expected'
+
+
+def _load_json(name):
+    return json.loads((INPUTS_DIR / name).read_text(encoding='utf-8'))
+
+
+def _load_text(name):
+    return (INPUTS_DIR / name).read_text(encoding='utf-8')
```
hu-border-collie: **[bug]** `_file_digest` normalizes `\r\n` → `\n` so `source_identity.file_digests` survive `core.autocrlf`, but `_load_text` does not. `build_fixture_plan` feeds `retrieval_blocks.txt` and `analysis_blocks.txt` straight into the canonical user prompt (lines 72–73), so those bytes are hashed into every `prompt_fingerprint` and into `plan_fingerprint`. The committed goldens were generated from LF text; a Windows checkout that converts those `.txt` files to CRLF will change the frozen fingerprints and fail `GoldenPlanTests`. `.gitattributes` does not pin `tests/fixtures/translation_plan_minimal/**` to `eol=lf`. JSON inputs are safe (`json.loads`); the multi-line reference texts are not. Macro is `.strip()`’d and is a single line, so it happens to be fine.

**Suggestion:** Apply the same `.replace('\r\n', '\n')` in `_load_text` (and/or add `tests/fixtures/translation_plan_minimal/** text eol=lf`). Add a test that a CRLF-normalized retrieval blob yields the frozen `prompt_fingerprint`.

tests/fixtures/translation_plan_minimal/inputs/file_jobs.json
```diff
@@ -0,0 +1,86 @@
+[
+  {
+    "file_rel_path": "chapter01/dialogue.rpy",
+    "file_path": "game/chapter01/dialogue.rpy",
+    "tasks": [
+      {
+        "id": "chapter01/dialogue.rpy:chapter01_start:1:2fc92d42",
+        "text": "Hey [Gil_name!t], did you finish the Dawn Chorus setlist?",
+        "line": 5,
```
hu-border-collie: **[suggestion]** README (lines 13–15) says each task `line` is the 0-based `TranslationUnit.line` for the matching source line. The adapter assigns `line` from `enumerate(lines)` (`engine_adapters/renpy.py`). In `game/chapter01/dialogue.rpy`, “Hey [Gil_name!t]…” is 1-based line 7 / 0-based 6, but `file_jobs` stores `line: 5` (and the rest of the dialogue tasks are similarly off-by-one: 6/7/8/12/13 instead of 6/7/8/9/13/14). `chapter02/strings.rpy` is correct (`old "New game"` is 0-based 4). The previous “re-derived as 5/6/7/8/12/13” pass looks like a missed fourth comment line on the dialogue file only. P1 does not parse the `.rpy`, so goldens stay internally consistent with `file_jobs`; P4 e2e that extracts from `game/` will not reproduce these `line_numbers` or this `plan_id`. Separately, fixture ids use `sha256(text)[:8]` (README) while production `build_identity_v2` uses `sha1[:8]`.

**Suggestion:** Set dialogue lines to 6/7/8/9/13/14, regenerate the expected plans, and add a test that every `file_jobs` `line` indexes a source line containing that task’s text. If `.rpy` remains narrative-only until P4, drop the “0-based source line” claim from the README rather than freezing the wrong numbers. Align id hashing with `build_identity_v2` if P4 is supposed to round-trip this fixture through the adapter.

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@
 - sync 预览、revision 预览/写回与 final review 统一消费同一套 quality finding schema：共享 normalize / validate / load / filter / digest 工具，各路径生成同一 `quality_findings.jsonl` + `quality_gate` 摘要；final review 仅做 LLM 语义字段到公共模型的适配，不冒充机械规则。
 - 同步初译增加局部前后文（`sync.context_before` / `sync.context_after`，默认 30/10，限定同文件与 translate block 边界）、`sync.macro_setting_file` 风格设定注入，以及不依赖 RAG 开关、不受 `top_k_terms` 截断的术语命中（`normalize_map` / 保留词 / 不可翻译词，全部实际命中进入提示词）；上下文构造事实与 macro 指纹写入 manifest 并纳入预览指纹。
 - LiteLLM 用户目录缓存不可写（只读 home、沙箱写入受限等）时自动回退到系统临时目录，并在 GUI 日志/状态栏提示缓存不会跨重启保留。
+- 新增执行策略无关的翻译计划纯核心 `translation_plan.py`（#346 P1）：D4 共享 chunking 与稳定 chunk/request ID、D1 block 边界局部上下文、D2 词法术语命中、六层上下文组装与预算诊断、D3 canonical prompt（Batch system/user 形态合并 Sync 独有规则）、plan/prompt/request 三级内容派生指纹与敏感键脱敏；`tests/fixtures/translation_plan_minimal` 冻结 sync 与 gemini_batch 的等价语义合同（同 chunk 的 `prompt_fingerprint` 一致）。Batch/Sync 的 chunking、hash key、局部上下文与词法术语函数改为委托共享实现。
```
hu-border-collie: **[suggestion]** The new bullet says Batch/Sync chunking, hash key, local context, and lexical-glossary functions were all switched to the shared implementation. What actually moved: batch `iter_translation_chunk_ranges` / `hash_key` / `task_text_char_count`, and sync `build_sync_local_context` / `retrieve_sync_glossary_hits`. Production sync still chunks with the inline `MAX_ITEMS=40` / `MAX_CHARS=12000` loop; batch local context is still raw `tasks[start-N:start]` with no block boundary; batch glossary hits are still `RAG_ENABLED`-gated. “六层上下文组装” is also ahead of the code: `CONTEXT_LAYER_BUDGET` is defined and mentioned in `assemble_context_layers` but never instantiated (five content layers + `dropped` / `total_char_used`).

**Suggestion:** Name the delegated functions explicitly and leave P2/P3 items (batch D1 window, batch D2 ungating, sync D4 loop, a real budget layer) out of the “已委托” sentence.

tests/test_translation_plan.py
```diff
@@ -0,0 +1,695 @@
...
+    kwargs = dict(
+        execution_strategy=strategy,
+        source_identity=_fixture_source_identity(),
+        config_snapshot=_load_json('config_snapshot.json'),
+        model_profile_snapshot=_load_json('model_profile.json'),
+        run_id='fixture-run',
+        # Shrink item budget so chapter01 splits across its two labels and the
+        # frozen golden plans exercise the D1 block boundary (the production
+        # D4 default 60/18000 is asserted in ChunkingTests).
+        chunk_policy=translation_plan.ChunkPolicy(max_items=4),
```
hu-border-collie: **[suggestion]** `ChunkPolicy(max_items=4)` splits chapter01 on the label boundary, which does freeze `block_bounded_*`. Each dialogue block is also ≤4 tasks, so both frozen chunks have `context_before_items` / `context_after_items` = 0 and the golden `user_prompt` never contains a non-empty `CONTEXT BEFORE`/`AFTER` rendered from task dicts. Canonical-prompt unit tests pass string lines (`'Previous line'`), not the dicts `build_local_context_window` actually returns. A regression in `_format_context_line` / speaker labels for dict tasks would not move the frozen `prompt_fingerprint`.

**Suggestion:** Use `max_items=3` (or add a fifth line to `chapter01_start`) so at least one golden request embeds intra-block context from task dicts, and assert the rendered `Gil:` / `Parker:` context lines in that `user_prompt`.

translation_plan.py
```diff
@@ -0,0 +1,1065 @@
...
+    combined_text = '\n'.join(
+        item.get('text', '') for item in target_items or [] if item.get('text')
+    )
+    if not combined_text:
+        return []
+    hits = []
+    seen = set()
+    # Sorted iteration: mappings with identical content but different
+    # insertion order must yield identical hit order (and prompts).
+    for source, target in sorted((normalize_map or {}).items()):
```
hu-border-collie, line 318: **[suggestion]** Sorting `normalize_map` items is the right determinism fix for plan identity, and the lexical “Existing glossary entries” block still lists every hit. Production `process_batch` also does `glossary_hits[:SYNC_RAG_TOP_K_TERMS]` for the RAG-on LOCKED TERMS list (`translator_runtime.py`, unchanged). That prefix is now the alphabetically first hits rather than insertion-order first, so with RAG enabled and `top_k` smaller than the hit count the LOCKED TERMS *set* can change, not just row order. CHANGELOG line 21’s “内容不变” is true for the always-injected lexical block, not for that truncated RAG list. Existing runtime tests `assertIn` both mappings and would not catch it.

**Suggestion:** Either document that the RAG `top_k` prefix is now key-sorted, or snapshot lexical hits separately from the truncated RAG list (e.g. do not reuse the sorted lexical sequence as the LOCKED TERMS ranking). A runtime test with two normalize hits and `top_k=1` should pin which term remains in LOCKED TERMS.

hu-border-collie: Addressed on `acd461a` by documenting the behavior rather than splitting the RAG ranking.

CHANGELOG now states that RAG `LOCKED TERMS` takes the key-sorted prefix, so when `top_k` < hit count the *set* can change (alphabetically-first hits kept). `test_sync_locked_terms_top_k_prefix_follows_sorted_hit_order` pins `top_k=1` keeping `encore` and dropping `setlist` from LOCKED TERMS while the lexical block still lists both.

tests/test_translation_plan.py
hu-border-collie: Fixed on `acd461a`.

Golden builds now use `ChunkPolicy(max_items=3)`. Frozen chapter01 chunks are `[0:3]` / `[3:6]`, with in-block `CONTEXT AFTER` / `CONTEXT BEFORE` rendered from task dicts as `Name (id)`. `test_golden_prompt_embeds_in_block_context_from_task_dicts` asserts those speaker-labeled lines, so a dict-rendering regression would move the frozen fingerprint.

tests/fixtures/translation_plan_minimal/inputs/file_jobs.json
hu-border-collie: Fixed on `acd461a`.

Dialogue `line` values are 6/7/8/9/13/14 and match the checked-in `.rpy` (0-based `TranslationUnit.line`). Task ids now use `sha1(text)[:8]` and match `build_identity_v2`. `test_file_jobs_lines_index_source_lines_containing_task_text` asserts every `file_jobs` line indexes a source line containing that task’s text.

tests/test_translation_plan.py
hu-border-collie: Fixed on `acd461a`.

`_load_text` now LF-normalizes, `_resolve_text_source` does the same for provider text (the bytes that actually enter `prompt_fingerprint`), and `.gitattributes` pins `tests/fixtures/translation_plan_minimal/**` to `eol=lf`. `test_crlf_normalization_keeps_frozen_fingerprints_stable` feeds CRLF retrieval/analysis blobs and asserts frozen fingerprints. Lock-manifest source digest for `.gitattributes` was refreshed; `verify_manifest()` matches.

CHANGELOG.md
hu-border-collie: Fixed on `acd461a`.

CHANGELOG now names the five actually-delegated functions (`iter_translation_chunk_ranges` / `hash_key` / `task_text_char_count`, `build_sync_local_context` / `retrieve_sync_glossary_hits`) and leaves the sync 40/12000 loop, batch raw-slice context, and batch RAG gate to P2/P3. The six-layer wording is scoped to five content layers plus assembly-level budget accounting.

tests/test_translation_plan.py
```diff
@@ -0,0 +1,770 @@
...
+        chunk_input = translation_plan.ChunkContextInput(
+            target_units=[],
+            analysis_blocks_text='',
+        )
+        assembly = translation_plan.assemble_context_layers(chunk_input)
+        empty_layers = [layer for layer in assembly.layers if not layer.text]
+        # Required/local/project render non-empty for real chunks; with no
+        # units and no providers the empty-text layers collapse to the first
+        # occurrence and later ones are dropped with a recorded reason.
+        self.assertLessEqual(len(assembly.layers), 5)
```
hu-border-collie: **[suggestion]** `test_duplicate_layer_texts_are_dropped_with_reason` cannot catch a regression that disables duplicate dropping. `empty_layers` is computed and never asserted; `assertLessEqual(len(assembly.layers), 5)` is always true (there are only five content layers); the `for dropped in assembly.dropped` loop is a no-op when `dropped` is empty. Removing the `seen_texts` skip would still pass this test.

**Suggestion:** Assert the drop actually happened, e.g. `self.assertTrue(assembly.dropped)`, `self.assertEqual({item['layer'] for item in assembly.dropped}, {retrieval, analysis})`, and `self.assertEqual(len(assembly.layers), 3)` (required/local/non-empty-or-first-empty-project). Use `empty_layers` or delete it.

translation_plan.py
```diff
@@ -0,0 +1,1072 @@
...
+        },
+    )
+
+
+def assemble_context_layers(chunk_input, context_policy=None):
+    """Assemble the five content layers in fixed rank order (issue #346).
+
+    Identical layer texts are dropped deterministically (later rank loses) and
+    recorded in the drop log; per-layer character budgets and truncation flags
+    feed the sixth (budget) layer's accounting.
```
hu-border-collie: **[suggestion]** After the CHANGELOG was corrected to five content layers plus assembly-level accounting, the code still describes a sixth budget layer: `assemble_context_layers` says truncation flags “feed the sixth (budget) layer's accounting”, the section header at line 358 still says “six layers”, and `CONTEXT_LAYER_BUDGET` / rank 6 are defined but never turned into a `ContextLayerResult`. That leftover design language will mislead P2/P3 wiring.

**Suggestion:** Drop `CONTEXT_LAYER_BUDGET` (or actually emit it). Rephrase the assembler docstring and section header to match the five content layers plus `total_char_used` / `dropped` accounting that the tests freeze.

translation_plan.py
```diff
@@ -0,0 +1,1072 @@
...
+* no optional SDK imports, no network access, no credential values — profile
+  snapshots pass through :func:`redact_sensitive` before entering a plan;
+* identical inputs rebuild byte-identical plans: every id and fingerprint is
+  content-derived via :func:`canonical_json`, never time- or randomness-based,
+  and unordered inputs (sets) are canonicalized before hashing;
+* ``run_id`` is an audit label only; retrieved content is captured by
+  ``prompt_fingerprint``, not by ``plan_id``;
+* retrieval and analysis layer texts are embedded in the canonical prompt
+  verbatim after budgeting, so those layers' accounting matches the prompt
+  byte-for-byte; the required/local/project layers are canonical renderings
```
hu-border-collie: **[suggestion]** The module contract still says retrieval/analysis texts are embedded “verbatim after budgeting, so those layers' accounting matches the prompt byte-for-byte”. `build_translation_plan` then does `layer.text.rstrip('\n')` and `'\n\n'.join(...)` (lines 997–1001), and `build_canonical_translation_user_prompt` may append another `\n\n`. `char_used` is `len(layer.text)` before that normalize, so trailing newlines in a budgeted slice are counted in assembly but not necessarily present 1:1 in `user_prompt`.

**Suggestion:** Either count `char_used` on the same post-rstrip strings that are joined into the prompt, or weaken the module docstring to “budgeted then joined on a fixed blank-line separator; `user_prompt` is authoritative.”

PR #388, opened 2026-08-23
#341: add first provider-neutral Embedding core slice

hu-border-collie:
## Summary

- add a dependency-free provider-neutral Embedding core
- define document/query semantics, backend/provider/model/dimension identity, deterministic serialization and fingerprints
- define validated batch request/result, usage metadata, timeout, stable error categories, and a minimal backend protocol
- reject credential-shaped metadata and invalid model/dimension/count/vector data at the core boundary
- recursively freeze validated metadata so credentials cannot be injected and fingerprints cannot drift after construction
- return field-level query/store compatibility diagnostics with an explicit rebuild_store action

## Scope and safety contract

This is the first independently mergeable core slice for #341. It does not import google-genai, LiteLLM, PySide6, or networking code. Store identities persist schema version, backend, provider, model, task type, output dimension, and a verified SHA-256 fingerprint. Identity deserialization requires that persisted fingerprint and fails closed on missing, blank, tampered, or invalid-version records.

Request/result/usage metadata recursively rejects credential-shaped keys and is frozen after validation. Public backend error strings come only from a closed category mapping and cannot carry raw provider bodies, URLs, headers, or error codes.

The compatibility check requires a document store identity and query request identity with matching backend, provider, model, and dimension. Mismatches must skip vector comparison and return deterministic rebuild diagnostics.

## Tests

- python -m unittest tests.test_embedding_backend -q — 22 passed
- isolated import smoke with google, litellm, and PySide6 imports blocked — passed
- python scripts/run_quality_gates.py lint-critical — passed
- Ruff on embedding_backend.py and tests/test_embedding_backend.py — passed
- git diff --check — passed

An earlier optional full CLI suite attempt was interrupted after an accidental duplicate run caused prolonged CPU contention; it is not reported as passing and is not required for this dependency-free, unconnected core slice. GitHub's separate cli-without-gui CI job passed on the original revision and will rerun for review changes.

## Not implemented

- Gemini adapter or migration of current embed_content calls
- OpenAI-compatible adapter
- runtime/configuration, doctor, GUI, or user-documentation wiring
- Sync RAG, Source Index, Published Project Analysis, or TranslationPlan production integration
- existing store migration or rewriting

## Design risk

OpenAI-compatible endpoints must receive a stable, non-secret provider/configuration identity. Reusing the same provider identity for distinct endpoints with the same model name could incorrectly describe different vector spaces as compatible. Query and document fingerprints intentionally differ by task type, so callers must use the field-level compatibility API rather than compare those fingerprints directly.

Refs #341

### DIFF
--- a/docs/plans/issue-341-provider-neutral-embedding-core.md
+++ b/docs/plans/issue-341-provider-neutral-embedding-core.md
@@ -1,13 +1,15 @@
 # Issue #341：Provider-neutral Embedding 纯核心
 
+状态：core v1 合同已冻结、尚未接入生产路径 · Issue #341 · 基线 `main@ed07a99`（2026-08-23）
+
 本阶段只冻结可独立合并的核心合同，不接入生产调用路径，也不改变现有 Gemini 行为。实现位于 `embedding_backend.py`，不依赖 Provider SDK、LiteLLM、PySide6、凭据或网络。
 
 ## 核心合同
 
-- `EmbeddingBackend` 是最小同步 protocol：adapter 接收一个已校验的 `EmbeddingBatchRequest`，返回 `EmbeddingBatchResult`，失败时映射为带稳定 `EmbeddingErrorCategory` 的 `EmbeddingBackendError`。
+- `EmbeddingBackend` 是最小同步 protocol：adapter 接收一个已校验的 `EmbeddingBatchRequest`，返回 `EmbeddingBatchResult`，失败时映射为带稳定 `EmbeddingErrorCategory` 的 `EmbeddingBackendError`。公开错误字符串按类别封闭映射，不携带 Provider 原始响应、URL、header 或错误码。
 - `EmbeddingTaskType` 只表达跨 Provider 的 `document` / `query` 语义。后续 Gemini adapter 负责映射到 `RETRIEVAL_DOCUMENT` / `RETRIEVAL_QUERY`；OpenAI-compatible adapter 可在 API 不支持 task 参数时仅保留该语义用于 identity。
 - `EmbeddingIdentity` 固定 `backend`、`provider`、`model`、`task_type` 和 `output_dimension`，并提供包含 schema version 的规范 JSON 与无凭据 SHA-256 fingerprint。store 应完整保存 document identity 的 `to_dict()`，不能只保存 model 名称。
-- batch request 固定 identity、输入顺序、有限正 timeout 和安全 metadata；result 绑定 request fingerprint，并校验向量数、每条维度、有限数值与 usage metadata。
+- batch request 固定 identity、输入顺序、有限正 timeout 和安全 metadata；metadata 会在入口校验后递归冻结，防止事后注入凭据或改变 fingerprint；result 绑定 request fingerprint，并校验向量数、每条维度、有限数值与 usage metadata。
 - metadata 中任何 credential-shaped key（包括嵌套的 API key、authorization、secret、password、credential ref、单数 token 等）都会在入口被拒绝，不会脱敏后继续参与持久化或指纹。
 
 ## Store / query 兼容性
--- a/docs/plans/issue-341-provider-neutral-embedding-core.md
+++ b/docs/plans/issue-341-provider-neutral-embedding-core.md
@@ -1,0 +1,30 @@
+# Issue #341：Provider-neutral Embedding 纯核心
+
+本阶段只冻结可独立合并的核心合同，不接入生产调用路径，也不改变现有 Gemini 行为。实现位于 `embedding_backend.py`，不依赖 Provider SDK、LiteLLM、PySide6、凭据或网络。
+
+## 核心合同
+
+- `EmbeddingBackend` 是最小同步 protocol：adapter 接收一个已校验的 `EmbeddingBatchRequest`，返回 `EmbeddingBatchResult`，失败时映射为带稳定 `EmbeddingErrorCategory` 的 `EmbeddingBackendError`。
+- `EmbeddingTaskType` 只表达跨 Provider 的 `document` / `query` 语义。后续 Gemini adapter 负责映射到 `RETRIEVAL_DOCUMENT` / `RETRIEVAL_QUERY`；OpenAI-compatible adapter 可在 API 不支持 task 参数时仅保留该语义用于 identity。
+- `EmbeddingIdentity` 固定 `backend`、`provider`、`model`、`task_type` 和 `output_dimension`，并提供包含 schema version 的规范 JSON 与无凭据 SHA-256 fingerprint。store 应完整保存 document identity 的 `to_dict()`，不能只保存 model 名称。
+- batch request 固定 identity、输入顺序、有限正 timeout 和安全 metadata；result 绑定 request fingerprint，并校验向量数、每条维度、有限数值与 usage metadata。
+- metadata 中任何 credential-shaped key（包括嵌套的 API key、authorization、secret、password、credential ref、单数 token 等）都会在入口被拒绝，不会脱敏后继续参与持久化或指纹。
+
+## Store / query 兼容性
+
+`check_store_query_compatibility(store_identity, query_identity)` 要求：
+
+1. store identity 的 task type 为 `document`；
+2. query identity 的 task type 为 `query`；
+3. backend、provider、model、output dimension 完全相同。
+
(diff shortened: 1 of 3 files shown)


embedding_backend.py
```diff
@@ -0,0 +1,523 @@
...
+            _non_negative_int_or_none(self.total_tokens, 'total_tokens'),
+        )
+        object.__setattr__(
+            self,
+            'metadata',
+            _validate_metadata_mapping(self.metadata, 'usage.metadata'),
+        )
+
+    def to_dict(self) -> dict[str, object]:
+        return {
```
hu-border-collie, line 353: **[bug]** `EmbeddingBatchRequest`, `EmbeddingUsage`, and `EmbeddingBatchResult` are frozen, and `inputs`/`vectors` are copied into tuples, but metadata is stored as a plain `dict` whose nested dicts/lists stay mutable. After a valid request is built, `request.metadata['api_key'] = 'secret'` succeeds, nested `request.metadata['nested']['api_key'] = 'injected'` succeeds, and `request.fingerprint` changes. That bypasses the recursive credential-shaped-key rejection this slice advertises and makes request/result fingerprints unstable for `validate_embedding_result`. The same hole exists at usage metadata (`embedding_backend.py:361`) and result metadata (`embedding_backend.py:414`).

**Suggestion:** Recursively freeze validated metadata (for example `MappingProxyType` for objects and tuples for arrays) before `__setattr__`, so post-init mutation raises and fingerprints stay bound to the checked payload. Add a unit test that `request.metadata['k'] = 'v'` and nested list/dict writes fail.

embedding_backend.py
```diff
@@ -0,0 +1,523 @@
...
+            'schema_version': EMBEDDING_REQUEST_SCHEMA_VERSION,
+            'identity': self.identity.to_dict(),
+            'inputs': list(self.inputs),
+            'timeout_seconds': self.timeout_seconds,
+            'metadata': dict(self.metadata),
+        }
+
```
hu-border-collie, line 321: **[suggestion]** `EmbeddingIdentity.from_dict` only verifies `fingerprint` when the key is present. A payload with edited `model`/`provider`/dimension and the fingerprint key removed loads successfully as a new identity. The plan says deserialization should verify the persisted fingerprint so fields cannot be rewritten while still standing in for an old store. Callers that persist `to_dict()` will include a fingerprint, but corrupt or hand-edited store records fail open. The checksum still cannot bind vectors, so this is fail-closed hygiene rather than a full authenticity guarantee.

**Suggestion:** Reject missing/blank `fingerprint` in `from_dict` (keep constructor/`to_dict` as the way to mint one). If incomplete records must be readable during a later migration slice, make that an explicit helper, not the default loader.

embedding_backend.py
```diff
@@ -0,0 +1,523 @@
...
+
+def _non_negative_int_or_none(value: object, field_name: str) -> int | None:
+    if value is None:
+        return None
+    if isinstance(value, bool) or not isinstance(value, int) or value < 0:
+        raise EmbeddingContractError(f'{field_name} must be a non-negative integer or null')
+    return value
+
+
```
hu-border-collie, line 128: **[suggestion]** `EmbeddingBackendError` stores the caller-supplied `message` (and `provider_code`) as the exception text. `SyncBackendError` in `sync_model_backend.py` deliberately maps every category to a closed safe string so raw provider bodies never land in logs or GUI diagnostics. Gemini/OpenAI embedding errors often include key or URL material; once this type is the frozen adapter contract, later doctor/GUI wiring will be tempted to display `str(exc)` directly.

**Suggestion:** Document that `message`/`provider_code` must already be non-secret, or mirror the sync backend and keep a category-only public string while putting detail behind an explicit safe summary helper.

embedding_backend.py
```diff
@@ -0,0 +1,523 @@
...
+        object.__setattr__(self, 'task_type', _task_type(self.task_type))
+        object.__setattr__(
+            self,
+            'output_dimension',
+            _positive_int(self.output_dimension, 'output_dimension'),
+        )
+
+    def identity_payload(self) -> dict[str, object]:
+        return {
+            'schema_version': EMBEDDING_IDENTITY_SCHEMA_VERSION,
```
hu-border-collie, line 241: **[suggestion]** The inline comment restates a type-narrowing workaround (`dict(normalized)` after the `Mapping` check) instead of a non-obvious constraint. Repo convention is that comments explain why, not what the next line does.

**Suggestion:** Drop the comment, or replace it with a one-line why if a later freeze (Issue 1) still needs an explicit copy.

embedding_backend.py
```diff
@@ -0,0 +1,523 @@
...
+            _required_text(text, f'inputs[{index}]')
+        timeout = self.timeout_seconds
+        if isinstance(timeout, bool) or not isinstance(timeout, (int, float)):
+            raise EmbeddingContractError('timeout_seconds must be a positive finite number')
+        timeout = float(timeout)
+        if not math.isfinite(timeout) or timeout <= 0:
+            raise EmbeddingContractError('timeout_seconds must be a positive finite number')
+        object.__setattr__(self, 'inputs', normalized_inputs)
+        object.__setattr__(self, 'timeout_seconds', timeout)
+        object.__setattr__(self, 'metadata', _validate_metadata_mapping(self.metadata))
```
hu-border-collie, line 311: **[nit]** `schema_version` is compared with `!= EMBEDDING_IDENTITY_SCHEMA_VERSION`. Because `True == 1` in Python, `{"schema_version": true, ...}` is accepted as v1. The rest of this module already rejects `bool` for integer fields via `_positive_int` / `_non_negative_int_or_none`.

**Suggestion:** Require `isinstance(version, int) and not isinstance(version, bool)` (or reuse `_positive_int`) before the equality check.

docs/plans/issue-341-provider-neutral-embedding-core.md
```diff
@@ -0,0 +1,30 @@
+# Issue #341：Provider-neutral Embedding 纯核心
```
hu-border-collie, line 1: **[nit]** `docs/plans/README.md` says plan docs are read by their opening status banner, and the other live plans in that directory have one. This freeze note has no status/baseline line, so it is easy to misread as a future sketch rather than the merged v1 contract.

**Suggestion:** Add a one-line banner (status, issue #341, baseline SHA, “frozen core / not wired”). Optionally index it from `docs/plans/README.md` in a follow-up; that file is outside this diff.

hu-border-collie: 已在 d7db6f6 补充文首状态横幅：core v1 合同已冻结、尚未接入生产路径，关联 Issue #341，并记录实现基线 main@ed07a99（2026-08-23）。未扩大修改到 plans 索引。

embedding_backend.py
hu-border-collie: 确认属实，已在 d7db6f6 改为 fail closed。EmbeddingIdentity.from_dict 现在要求 fingerprint 必须是非空字符串，并在构造字段后严格校验相等；缺失、空白和篡改均拒绝。新增 test_from_dict_requires_persisted_fingerprint，原篡改测试继续通过。

embedding_backend.py
hu-border-collie: 确认这是安全边界问题，已在 d7db6f6 采用封闭公开字符串。EmbeddingBackendError 构造器不再接收或保存原始 message/provider_code，只保留稳定 enum category 与 retryable；str(exc) 来自 _SAFE_ERROR_MESSAGES。新增测试验证公开文本稳定且不含原 quota 细节，设计文档也明确禁止附加 Provider body、URL、header 或 code。

embedding_backend.py
hu-border-collie: 确认属实，已在 d7db6f6 修复。_validate_metadata 现在把对象递归冻结为 MappingProxyType、数组冻结为 tuple；request/usage/result 的 to_dict 通过 _metadata_to_json 返回脱离内部状态的 JSON 副本。新增 test_metadata_is_recursively_immutable_and_fingerprint_stays_stable 与 test_usage_and_result_metadata_are_recursively_immutable，覆盖顶层、嵌套 dict/list 写入、外部源对象事后修改及 fingerprint 稳定性。

embedding_backend.py
hu-border-collie: 已在 d7db6f6 处理。原先复述类型收窄的注释已删除；该 helper 现在直接返回递归冻结后的 Mapping，非预期分支仅保留防御性合同错误。

embedding_backend.py
hu-border-collie: 确认属实，已在 d7db6f6 显式拒绝 bool、非 int 和非 v1 的 schema_version。新增 test_from_dict_rejects_boolean_schema_version，覆盖 JSON true 不能冒充版本 1。

PR #390, opened 2026-08-23
feat(plan): Gemini Batch consumes TranslationPlan (#346 P2)

hu-border-collie:
## Summary

Implements **#346 P2**: the Gemini Batch translation build path now consumes `TranslationPlan` instead of assembling its own chunks/context/prompts.

## Changes

- `gemini_translate_batch.build_chunks` now delegates to `translation_plan.build_translation_plan` and keeps the legacy chunk shape for `submit` / `probe` / `split` / `repair` / cost-estimate compatibility.
- `build_batch_request` only wraps planned `TranslationRequest`s in the Gemini Batch envelope, and writes `request_id` / `plan_id` / `chunk_id` / `prompt_fingerprint` / `request_fingerprint` into each `requests.jsonl` row.
- `create_batch_package` writes a `translation_plan` block into manifest v2.
- Lexical glossary (`normalize_map` / `preserve_terms` / `non_translatable_exact`) is no longer gated by `RAG_ENABLED` nor truncated by `RAG_TOP_K_TERMS` (D2).
- `batch_cost_estimate` prefers `translation_plan.chunks` when present, with legacy manifest fallback.
- Retry subchunk creation drops stale plan-rendered fields so retry packages regenerate a legacy prompt instead of reusing a stale request.
- Updated the golden batch request snapshot; only the canonical system/user prompt hashes changed.

## Tests

- Targeted: `test_translation_plan test_batch_rag test_batch_repair test_batch_golden_corpus test_batch_submit_recovery test_source_index test_batch_cost_estimate test_identity_v2 test_translation_core test_model_profile test_translation_ab_experiment` → **338 tests OK (skipped=1)**
- Full CLI suite in this sandbox with user-site-packages disabled to avoid the unrelated local `tests` shadowing package:
  `PYTHONPATH=tests PYTHONNOUSERSITE=1 python3 -B tests/run_cli_tests.py -q` → **1475 tests OK (skipped=12)**
- `git diff --check` passes.

## Notes

- `PYTHONNOUSERSITE=1` is only needed in this sandbox because `ebmlite` installs a top-level `tests` package that shadows the repo's local `tests/` directory; it is unrelated to this PR.

### DIFF
--- a/batch_cost_estimate.py
+++ b/batch_cost_estimate.py
@@ -208,7 +208,17 @@ def estimate_manifest_tokens(manifest, pricing_config=None):
     max_output_tokens = int(settings.get('max_output_tokens') or 0)
 
     summary = manifest.get('summary') if isinstance(manifest.get('summary'), dict) else {}
-    chunk_count = int(summary.get('chunk_count') or 0)
+    translation_plan = manifest.get('translation_plan') or {}
+    planned_chunks = (
+        translation_plan.get('chunks')
+        if isinstance(translation_plan, dict)
+        else None
+    )
+    chunk_count = (
+        int(len(planned_chunks))
+        if isinstance(planned_chunks, list) and planned_chunks
+        else int(summary.get('chunk_count') or 0)
+    )
     if chunk_count <= 0 and isinstance(manifest.get('chunks'), list):
         chunk_count = len(manifest['chunks'])
     # final-review and other request-only packages expose unit_count / request_count
--- a/batch_cost_estimate.py
+++ b/batch_cost_estimate.py
@@ -208,17 +208,7 @@ def estimate_manifest_tokens(manifest, pricing_config=None):
     max_output_tokens = int(settings.get('max_output_tokens') or 0)
 
     summary = manifest.get('summary') if isinstance(manifest.get('summary'), dict) else {}
-    translation_plan = manifest.get('translation_plan') or {}
-    planned_chunks = (
-        translation_plan.get('chunks')
-        if isinstance(translation_plan, dict)
-        else None
-    )
-    chunk_count = (
-        int(len(planned_chunks))
-        if isinstance(planned_chunks, list) and planned_chunks
-        else int(summary.get('chunk_count') or 0)
-    )
+    chunk_count = int(summary.get('chunk_count') or 0)
     if chunk_count <= 0 and isinstance(manifest.get('chunks'), list):
         chunk_count = len(manifest['chunks'])
     # final-review and other request-only packages expose unit_count / request_count
--- a/gemini_translate_batch.py
+++ b/gemini_translate_batch.py
@@ -1859,6 +1859,34 @@ def retrieve_glossary_hits(target_items):
     )
 
 
+def retrieve_revision_glossary_hits(target_items):
+    """Return the legacy RAG-gated, top-k-capped glossary hits for revision.
+
+    Revision packages keep their pre-#346 contract: no RAG means no LOCKED
+    TERMS, and RAG top-k still caps the reference list. Only the translation
+    plan path (issue #346 P2/D2) deliberately uses the unbounded lexical
+    helper above.
+    """
+    if not RAG_ENABLED:
+        return []
+    combined_text = '\n'.join(item.get('text', '') for item in target_items if item.get('text'))
+    if not combined_text:
+        return []
+    hits = []
+    seen = set()
+    for source, target in (legacy.NORMALIZE_TRANSLATION_MAP or {}).items():
+        if source and source in combined_text and source not in seen:
+            hits.append({'source': source, 'target': target, 'kind': 'normalize'})
+            seen.add(source)
+    for term in legacy.PRESERVE_TERMS:
+        if not isinstance(term, str) or not term.strip():
+            continue
+        if term in combined_text and term not in seen:
+            hits.append({'source': term, 'target': term, 'kind': 'preserve'})
+            seen.add(term)
+    return hits[:RAG_TOP_K_TERMS]
+
+
 def format_glossary_hits_block(hits, empty_label='(none)'):
     return prompt_context.format_glossary_hits_block(hits, empty_label)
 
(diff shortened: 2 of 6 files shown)


batch_cost_estimate.py
```diff
@@ -208,7 +208,17 @@ def estimate_manifest_tokens(manifest, pricing_config=None):
...
 
     summary = manifest.get('summary') if isinstance(manifest.get('summary'), dict) else {}
-    chunk_count = int(summary.get('chunk_count') or 0)
+    translation_plan = manifest.get('translation_plan') or {}
+    planned_chunks = (
+        translation_plan.get('chunks')
+        if isinstance(translation_plan, dict)
+        else None
+    )
+    chunk_count = (
```
hu-border-collie: **[bug]** `estimate_manifest_tokens` now prefers `len(manifest['translation_plan']['chunks'])` over `summary.chunk_count` whenever that list is non-empty. `copy_split_context_metadata` copies the *parent* `translation_plan` wholesale into every split part and every retry package (`gemini_translate_batch.py:2838`), while those packages set `summary.chunk_count` to the part/retry size. Submit then recomputes the estimate via `ensure_manifest_cost_estimate` if `cost_estimate` is missing (split/retry do not attach one). Result: a 5-chunk retry of a 400-chunk job is billed as 400 * `max_output_tokens`, and `--max-cost` can block a cheap split part because it still sees the parent plan. On the original package the two counts are equal, so preferring the plan never helps and only goes stale after split/retry. Legacy packages without `translation_plan` still fall back correctly.

**Suggestion:** Estimate output tokens from this package's actual work (`manifest['chunks']`, `summary.chunk_count`, or the JSONL `request_count` already parsed in this function). If the parent plan is kept on split/retry for identity, filter `translation_plan.chunks` / `request_summaries` to the part's `chunk_id`s, or ignore `translation_plan.chunks` whenever `split_from_manifest` / `retry_of_manifest` is set. Add a test with a 2-chunk package whose copied `translation_plan.chunks` has length 10.

gemini_translate_batch.py
```diff
@@ -5930,6 +6181,26 @@ def build_retry_subchunk(chunk, start, end, sub_index):
         context_future = (copy.deepcopy(items[end:min(len(items), end + BATCH_CONTEXT_AFTER)]) + context_future)[:BATCH_CONTEXT_AFTER]
     subchunk['context_past'] = context_past
     subchunk['context_future'] = context_future
+    # The subchunk has new items and a new key; a P2 plan request from the
+    # parent would no longer match. Drop plan-rendered fields so
+    # build_batch_request regenerates a legacy prompt for the subchunk instead
+    # of reusing a stale TranslationRequest.
+    for plan_field in (
```
hu-border-collie: **[bug]** Dropping plan-rendered fields on retry subchunks avoids reusing a stale parent `TranslationRequest`, but `_chunk_has_plan_request` then falls through to the legacy builders (`build_system_instruction` / `build_user_prompt` at `gemini_translate_batch.py:3241`). Those call `translation_core.build_translation_system_instruction` / `build_translation_user_prompt`, not the canonical P1 contract the initial package just sent. Material drift vs the plan path: (1) system instruction loses “Keep all person names in English” and “No markdown, no Pinyin”; (2) glossary is re-rendered as `LOCKED TERMS` via `format_glossary_hits_block`, not `Existing glossary entries` / `render_lexical_glossary_text`; (3) `non_translatable_exact` hits now stored on the parent (`target: ''`) render as `- Term -> ` instead of `Non-translatable:`; (4) subchunk context is a raw `BATCH_CONTEXT_*` slice, not D1 block-bounded windows. Full-chunk retries that do not split keep the parent request (correct); any split or item-scoped retry does not. P2's stated contract is that submit/probe/split/repair consume the plan request rather than regenerating a prompt.

**Suggestion:** Rebuild a derived `TranslationRequest` for the subchunk (same providers, D7-style lineage id) and keep wrapping it in `build_batch_request`. If a legacy fallback must remain for old manifests, gate it on missing plan fields on the *parent*, not on every new subchunk. Assert in tests that a split retry's system instruction still contains the canonical name/pinyin rules and that a `non_translatable` hit still renders as `Non-translatable:`.

gemini_translate_batch.py
```diff
@@ -3233,71 +3287,261 @@ def count_translation_chunks(file_jobs):
...
+
+def _render_batch_retrieval_reference_text(history_hits, story_hits, source_hits):
+    """Render the retrieval layer without the lexical glossary block.
+
+    Lexical glossary is injected by the canonical prompt's project layer
+    (issue #346, D2); the retrieval layer carries only RAG history, source
+    index, and story memory reference text.
+    """
+    blocks = []
+    if history_hits:
```
hu-border-collie, line 3400: **[bug]** `_render_batch_retrieval_reference_text` concatenates RAG history, Source Index, and Story Memory into one retrieval blob. `ContextPolicy` then budgets that layer at `history_char_limit + story_char_limit` (220+1200=1420; `gemini_translate_batch.py:3475`). `_retrieval_layer` truncates to that cap. Pre-P2 Batch injected those sections independently (history per-hit 220, source `TOP_K * CHAR_LIMIT`, story 1200) with no combined 1420 envelope. Any two retrieval sources enabled at once now silently clip later sections: default Source Index budget is already ~880 chars of excerpt plus headers, so Story Memory is cut in half or dropped; RAG+story overruns even with Source Index off. The same pattern hits analysis: `_render_batch_analysis_reference_text` concatenates PROJECT BRIEF + PROJECT LOCAL CONTEXT, then `analysis_char_limit=PROJECT_ANALYSIS_MAX_BRIEF_CHARS` (4000; `gemini_translate_batch.py:3478`). A brief already at `max_brief_chars` consumes the whole layer (headers included) and local label/route context never reaches the model.

**Suggestion:** Either raise the retrieval backstop by the Source Index char budget (and the analysis backstop by local-context budgets) before calling `build_translation_plan`, or pre-trim each section to its own limit and pass already-fitting text so the D5 backstop only catches true overflow. Golden/unit coverage should enable story+source (and a max-length brief+local) and assert both section headers survive in `user_prompt`.

tests/test_batch_rag.py
```diff
@@ -190,6 +190,126 @@ def test_build_chunks_respects_source_char_limit(self):
...
+            batch_mod.BATCH_TARGET_CHARS = old_values['target_chars']
+            batch_mod.BATCH_CONTEXT_BEFORE = old_values['context_before']
+            batch_mod.BATCH_CONTEXT_AFTER = old_values['context_after']
+            batch_mod.RAG_ENABLED = old_values['rag_enabled']
+            batch_mod.STORY_MEMORY_ENABLED = old_values['story_enabled']
+            batch_mod.legacy.PRESERVE_TERMS = old_preserve_terms
+
+        self.assertTrue(chunks)
+        hits = chunks[0].get('glossary_hits') or []
+        self.assertTrue(any(str(hit.get('source') or '') == 'Dawn Chorus' for hit in hits))
```
hu-border-collie, line 238: **[suggestion]** New tests check `glossary_hits` on the chunk and that `build_batch_request` copies plan fields, but they never assert the D2 contract the changelog claims: with `RAG_ENABLED=False`, the *prompt* contains the hit (`Existing glossary entries` / `Preserve: Dawn Chorus`). They also miss `normalize_map` / `non_translatable_exact`, `create_batch_package` writing `translation_plan`, retry subchunks dropping plan fields, and `estimate_manifest_tokens` preferring `translation_plan.chunks`. The golden `request_snapshot` only hashes system/user text and `request.keys()` inside the Gemini envelope, so jsonl `request_id` / fingerprints and manifest `translation_plan` are unpinned.

**Suggestion:** Assert glossary text in `chunk['user_prompt']` (and that it is absent from a retrieval `LOCKED TERMS` block). Add a retry-subchunk test that the rebuilt request is canonical, and a cost-estimate test with a stale parent plan. Optionally extend `_request_snapshot` / `_manifest_snapshot` with `request_id` and `plan_fingerprint`.

gemini_translate_batch.py
```diff
@@ -1843,24 +1844,19 @@ def embed_query_text(query_text):
 
 
 def retrieve_glossary_hits(target_items):
```
hu-border-collie, line 1846: **[suggestion]** `retrieve_glossary_hits` is now the unbounded D2 helper (no RAG gate, no `RAG_TOP_K_TERMS`, includes `non_translatable_exact`). `build_revision_chunks` still calls it as `retrieve_glossary_hits(target_items) if RAG_ENABLED else []` (`gemini_translate_batch.py:4639`). Revision stays RAG-gated, but with RAG on it now injects every hit including non-translatable terms and no longer truncates to `top_k_terms`. That is an unintended revision-package behavior change in a translation-plan PR.

**Suggestion:** Keep a revision-specific helper (or pass `top_k`) if revision should retain the old LOCKED TERMS cap, or explicitly adopt D2 for revision and test it. Do not share the unbounded translation helper unless that change is intended.

PR #394, opened 2026-08-23
feat(sync): consume TranslationPlan (#346 P3)

hu-border-collie:
Refs #346

## Summary

Implements **#346 P3**: ordinary synchronous initial translation now consumes the shared `TranslationPlan` introduced by P1 and already used by Gemini Batch in P2.

This PR keeps `TranslationPlan` schema v1 intact. It does not implement P4/P5, #347's durable executor, or any Source Index / Published Project Analysis / Embedding adapter work.

## What changed

- Sync builds one shared plan for the full pending-task snapshot and executes its fixed `PlanChunk` / `TranslationRequest` sequence instead of maintaining separate inline chunk/context/prompt semantics in `translator_runtime.py`.
- Sync and Batch use the same canonical system/user prompt, response schema, local-context algorithm, lexical glossary injection, `ContextAssembler` content, stable request identity, and prompt/request/plan fingerprints.
- Gemini Sync passes canonical system instruction through Google GenAI config. LiteLLM and custom OpenAI-compatible providers receive the same instruction through `config["system_instruction"]`, which the existing adapter converts to a system message.
- Invalid/incomplete responses preserve D7: the initial plan is immutable; targeted retries use deterministic `--T` child requests and binary splits use deterministic `--L` / `--R` children. Parent/child request IDs and reasons are recorded in `model_contract.retry_lineage`.
- Sync preview manifests bind `translation_plan`, `plan_fingerprint`, `request_ids`, per-file source digests, and adapter identity into the existing preview fingerprint and apply-time validation. Existing source snapshot, structure, quality, preview/apply, and operation-identity checks remain authoritative.
- Legacy Sync preview manifests without `translation_plan` remain readable/applicable through the existing legacy validation path.
- Macro and lexical glossary context remains present when RAG is disabled.

## D4 default change and mitigations

The default Sync chunk policy changes from **40 items / 12,000 source characters** to the frozen D4 choice of **60 items / 18,000 source characters**, matching ordinary Gemini Batch.

Two mitigations are included:

1. The selected `ModelCapabilities.context_budget_tokens`, when known, is checked before model calls; an oversized planned request fails preflight with the request ID, estimate, and model budget.
2. Existing explicit `sync.chunk_size` and `sync.max_source_chars` configuration remains supported, so projects that need smaller requests can retain their previous values.

The D5 Sync Story Memory default is also aligned from 800 to 1,200 characters. GUI defaults, `translator_config.example.json`, setup/sync docs, and tests are synchronized.

## Stable equivalence and explainable differences

Production-adapter tests build Sync and Gemini Batch requests from the same fixture and assert equality of canonical system/user prompt, response schema, expected IDs, and all ContextAssembler layer content usage/truncation facts. Sync plan identity is stable across different runtime run IDs.

Transport and generation differences remain explicit and auditable: Sync uses `sync_stage=initial_translation`, timeout, and its output-token limit; Batch uses `batch_key`, remote envelope metadata, and its own output-token/thinking configuration. These differences produce explainably different request fingerprints without changing the model-facing semantic contract.

## Safety and compatibility

- No P1/P2 schema v1 fields were changed or reinterpreted.
- No `check -> apply` or Sync preview/apply guard was bypassed.
- Plan/source/writeback identity is revalidated before the first apply write.
- Existing explicit Sync configuration remains accepted.
- Legacy preview manifests remain supported.
- No credentials or sensitive headers enter plan/diagnostic artifacts.

## Verification

Fresh on head `cfc665f` against unchanged `origin/main@a7319aa`:

- Targeted TranslationPlan/core/CLI-contract/Sync backend/runtime/preview/GUI-settings suite: **362 passed**
- `git diff --check origin/main...HEAD`: **clean**

Previously completed on the same head and unchanged base:

- `python -B tests/run_cli_tests.py -q`: **1507 passed, 1 skipped**
- `python -B tests/run_gui_tests.py -q`: **1155 passed**
- `python scripts/run_quality_gates.py all`: **all blocking gates passed**

## Out of scope / follow-up

- P4: broader golden/plan-diff diagnostics and remaining safety/diagnostic consolidation.
- P5: #341 Source Index, Published Project Analysis, and provider-neutral Embedding integration.
- #347: durable checkpoints, leases, cancellation, concurrency, and resume execution.


translation_plan.py
```diff
@@ -1079,3 +1079,116 @@ def build_translation_plan(
...
+    )
+    chunk_input = ChunkContextInput(
+        file_rel_path=str(file_rel_path or ''),
+        target_items=items,
+        target_units=units,
+        context_window=context_window or translation_core.ContextWindow(),
+        local_context_diagnostics=dict(local_context_diagnostics or {}),
+        macro_setting=str(macro_setting or ''),
+        lexical_glossary_hits=lexical_hits,
+        retrieval_blocks_text=str(retrieval_blocks_text or ''),
```
hu-border-collie: **[bug]** `derive_translation_request` copies `retrieval_blocks_text` / `analysis_blocks_text` into `ChunkContextInput` without the LF normalization `build_translation_plan` applies through `_resolve_text_source`. Sync captures also store the provider return value before that replace (`translator_runtime.py` retrieval_provider). Parent requests therefore embed LF-normalized retrieval, while `--T`/`--L`/`--R` children re-assemble from the raw capture. On Windows, CRLF in Story Memory or RAG hit text makes retry prompts (and their prompt fingerprints) diverge from the frozen parent contract that D7 claims to preserve.

**Suggestion:** Normalize the same way as `_resolve_text_source` (at least `str(...).replace('\r\n', '\n')`) inside `derive_translation_request`, and store the already-normalized retrieval text in the Sync captures so parent and child see identical bytes.

translator_runtime.py
```diff
@@ -4670,40 +4858,45 @@ def process_batch(
...
+            batch,
+            glossary_hits=locked_glossary_hits,
+            history_hits=history_hits,
+            story_hits=story_hits,
+            context_window=context_window,
+            normalize_map=normalize_map,
+            non_translatable_terms=non_translatable_terms,
+        )
+    else:
+        prompt = translation_request.user_prompt
```
hu-border-collie, line 5032: **[suggestion]** Plan-backed `process_batch` never prints `Sync RAG memory hits`. That log now lives only in the `translation_request is None` legacy branch, but ordinary initial translation always passes a plan request. RAG is still queried (at plan-build time) and `rag_stats` still land in preview `prompt_context`, so this is operator-visible stdout only, not a missing retrieval.

**Suggestion:** Log hit counts while building the plan (once per request, from each capture's `rag_stats`) or keep a single summary after `build_sync_translation_plan`.

sync_translation_preview.py
```diff
@@ -20,12 +20,77 @@
...
+            and str(entry.get("source_sha256") or "")
+            != str(file_digests.get(relative_path) or "")
+        ):
+            raise ValueError(
+                "Sync preview source digest does not match its TranslationPlan."
+            )
+        writeback_plan = entry.get("writeback_plan")
+        if not isinstance(writeback_plan, dict):
+            continue
+        # WritebackPlan project/source fingerprints are scoped differently:
```
hu-border-collie: **[suggestion]** The writeback-binding block is a four-line design-rationale comment (why project/source fingerprints are not compared, why only engine/adapter_version are). It restates architecture history rather than a non-obvious constraint on the next line.

**Suggestion:** Drop it, or keep one short WHY note such as “WritebackPlan source fingerprints are per-file live snapshots; bind engine/adapter_version only.”

translator_runtime.py
```diff
@@ -5188,9 +5477,11 @@ def process_batch_with_retry(
             contract_diagnostics=contract_diagnostics,
             contract_failures=contract_failures,
             retry_kind='split_retry',
-            # See the split-left call above: context window is not reused.
+            # See the split-left call above.
```
hu-border-collie: **[nit]** `# See the split-left call above.` is leftover from the old “context window is not reused” comments. The left-hand call now only notes that D7 children keep the original window, so this right-hand pointer explains nothing.

**Suggestion:** Delete it.

translator_runtime.py
hu-border-collie: 已在 47dcbff 修复。plan-backed 路径现在在 retrieval_provider 构建每个 request 时输出 Sync RAG memory hits，rag_stats 仍继续进入 preview diagnostics。新增 operator stdout 回归测试。

translation_plan.py
hu-border-collie: 已在 47dcbff 修复。新增统一的 normalize_context_provider_text，build_translation_plan 与 derive_translation_request 走同一 LF 规范化；Sync capture 也保存规范化后的 retrieval 文本。回归测试用 CRLF retrieval/analysis 构建派生请求，并断言其完整序列化结果与 LF 输入一致。

translator_runtime.py
hu-border-collie: 已在 47dcbff 删除这条无信息的 split-right 指针注释。

sync_translation_preview.py
hu-border-collie: 已按建议在 47dcbff 收敛为一行 WHY 注释：WritebackPlan source fingerprints are per-file live snapshots。

PR #406, opened 2026-08-31
feat: export offline quality review reports

hu-border-collie:
Summary: adds a shared quality-report exporter, a machine-readable CLI command, and a GUI export action. The generated single-file HTML summarizes warnings, blockers, acknowledgement state, rules, and files; supports local search, filters, responsive layout, print, and reduced motion; escapes finding content; loads no external resources; and refuses to overwrite the manifest or findings source. Documentation and diagnostics command references are updated. Validation: 140 focused quality-report, GUI-report, diagnostics, user-copy, and CLI-contract tests pass; critical Ruff gate passes; git diff --check passes. Additional full-suite attempts exposed unrelated local baseline failures: six existing GUI project-gate assertions and eight translation A/B routing tests caused by an unprefixed LiteLLM model profile.


gemini_translate_batch.py
```diff
@@ -19103,6 +19131,22 @@ def _run_usage_command():
     if command == 'check':
         return check_results(args.target or None)
 
+    if command == 'quality-report':
+        manifest = load_manifest(args.target or None)
+        result = quality_report_export.export_quality_report(
```
hu-border-collie: **[bug]** `quality-report` is advertised as a `machine_output` command, but expected failures from `export_quality_report` (`QualityReportExportError`, a `ValueError`) are not converted to `cli_contract.MachineContractError`. Sibling `quality-ack` already maps missing/invalid JSONL to stable codes such as `QUALITY_FINDINGS_UNAVAILABLE` and `INVALID_QUALITY_FINDINGS_JSON`. Here, a missing report, bad JSONL, or protected-path clash dumps a traceback in text mode; JSON mode goes through the generic `except Exception` path, prints a traceback on stderr, and emits `error.code=INTERNAL_ERROR`. With `--strict-exit-codes`, `classify_error()` also misses the Chinese messages (they do not contain English markers like `not found`), so agents still get an internal-error exit instead of a precondition/usage code.

**Suggestion:** Catch `QualityReportExportError` in the CLI handler (or reuse `read_quality_findings()` / the existing path resolver) and raise `MachineContractError` with stable `code_name` values (`QUALITY_FINDINGS_UNAVAILABLE`, `INVALID_QUALITY_FINDINGS_JSON`, and a dedicated overwrite/path code). Keep `QualityReportExportError` for the GUI, which already displays `str(exc)`.

gemini_translate_batch.py
```diff
@@ -20236,6 +20299,11 @@ def add_path(value):
             os.path.join(output_dir, engine_reuse.DEFAULT_CANDIDATES_FILENAME)
         )
 
+    # Keep the JSON envelope separate from the generated HTML artifact.
+    html_output = getattr(args, 'html', None)
+    if isinstance(html_output, str) and html_output.strip():
```
hu-border-collie, line 20326: **[bug]** `--output-file` is only protected against an *explicit* `--html` path. The default HTML destination is `dirname(quality_findings.jsonl)/quality_report.html` and is not added to the protected set. Preflight therefore allows `--output json --output-file <package>/quality_report.html` without `--html`. Dispatch writes the HTML first, then the JSON envelope atomically replaces that same path, so `artifacts.quality_report` points at a JSON file and the HTML report is gone. That is a realistic mix-up for a command named `quality-report` whose default artifact is `quality_report.html`. The new test only covers the both-flags-explicit case.

**Suggestion:** When `command == 'quality-report'`, resolve the same output path `export_quality_report` will use (explicit `--html`, else findings-dir/`quality_report.html`) and pass it to `add_path()` during `_collect_output_file_protected_paths`. Extend the CLI-contract test to omit `--html` and still expect `OUTPUT_FILE_PATH_CONFLICT`.

quality_report_export.py
```diff
@@ -0,0 +1,342 @@
...
+    if resolved_output:
+        resolved_output = os.path.abspath(resolved_output)
+    else:
+        resolved_output = os.path.join(
+            os.path.dirname(os.path.abspath(source_path)),
+            DEFAULT_REPORT_FILENAME,
+        )
+    if os.path.isdir(resolved_output):
+        raise QualityReportExportError(f"HTML 输出路径不能是目录：{resolved_output}")
+    protected_paths = {
```
hu-border-collie: **[suggestion]** Overwrite protection only blocks the current manifest and the findings JSONL. `--html` (and the GUI save dialog) can still replace other task files that `_collect_output_file_protected_paths` already treats as sacred for JSON envelopes, notably `results.jsonl`, `check_failures.jsonl`, and applied `.rpy` paths. That is a new write path that does not reuse the existing task-path guard.

**Suggestion:** Reuse the manifest protected-path set (or at least `results.jsonl` / `check_failures.jsonl`) inside `export_quality_report` before `atomic_write_text`, so CLI `--html` and the GUI picker share the same refusal.

quality_report_export.py
```diff
@@ -0,0 +1,342 @@
...
+        ("证据", "evidence"),
+        ("建议", "suggestion"),
+    ):
+        value = finding.get(field)
+        if value not in (None, ""):
+            detail_rows.append(
+                f'<div class="detail"><dt>{label}</dt><dd>{_escape(value)}</dd></div>'
+            )
+
+    return (
```
hu-border-collie, line 198: **[suggestion]** Each finding card shows file/line/`item_id` but never `finding_id`. Offline review of this HTML cannot be turned into `quality-ack --finding <id>` without reopening the JSONL or GUI list. The card already has a metadata row (`<code>` currently repeats `reason_code`).

**Suggestion:** Include the escaped `finding_id` in the card (and in `data-search`) so the offline report is usable with the existing acknowledgement CLI.

quality_report_export.py
hu-border-collie: 已在 ed3e9e7 修复：公开并复用了完整 manifest 保护路径集合，CLI 与 GUI 都把该集合传给 export_quality_report；导出会拒绝覆盖 results.jsonl、check_failures.jsonl、manifest 文件目标及已应用脚本等任务路径。新增了集合覆盖和不覆盖内容的测试。

quality_report_export.py
hu-border-collie: 已在 ed3e9e7 修复：每张卡片现在显示转义后的 finding_id，并把 finding_id 纳入 data-search；报告可直接复制该 ID 配合 quality-ack 使用。

gemini_translate_batch.py
hu-border-collie: 已在 ed3e9e7 修复：导出器错误现在携带稳定 code/details，CLI 会转换为 MachineContractError。缺失/不可读映射为 QUALITY_FINDINGS_UNAVAILABLE，坏 JSONL 映射为 INVALID_QUALITY_FINDINGS_JSON，路径冲突与写入失败也有专用错误码；补了机器模式错误契约测试。

gemini_translate_batch.py
hu-border-collie: 已在 ed3e9e7 修复：预检现在通过与导出器共用的 resolve_report_output_path 解析默认 quality_report.html，并将其加入 --output-file 保护集合。新增了省略 --html 时仍拒绝冲突的回归测试。

PR #407, opened 2026-09-02
feat(engine): add Tyrano native catalog writeback planning

hu-border-collie:
## 摘要

- 为公共写回合同新增 `json_catalog_set` 语义化 JSON catalog 操作，同时保持既有 `text_span_replace` 计划兼容
- 为 TyranoScript V600+ adapter 实现译文验证、原生 lang JSON 写回计划、重复 row 合并与冲突拒绝
- 将 catalog 纳入 source snapshot，并对缺 row、快照变化、空译文、控制字符和非法 JSON fail closed
- 保留 UTF-8 BOM、CRLF 与 JSON 键顺序；写回渲染不会修改 `.ks`
- 同步持久化反序列化、现行文档与回归测试

## 验证

- `python -m unittest tests.test_engine_adapter_p5_tyrano tests.test_engine_adapter_p2 tests.test_sync_translation_preview -q`（72 tests）
- `python -m unittest tests.test_tyranoscript_p5_fixtures -q`（17 tests）
- `git diff --check`

本地全量 CLI runner 共执行 1806 tests，出现 2 个既有 ModelProfile 顺序相关错误；两个失败用例分别单独复跑均通过。工作区的未跟踪 `tests/test_issue_341_acceptance.py` 未包含在本 PR。

Refs #399
Refs #265

### DIFF
--- a/engine_adapters/writeback.py
+++ b/engine_adapters/writeback.py
@@ -364,6 +364,8 @@ def render_writeback_plan(
         source_text = document.text()
         newline = "\r\n" if "\r\n" in source_text else "\n"
         serialized = json.dumps(data, ensure_ascii=False, indent=2)
+        if newline != "\n":
+            serialized = serialized.replace("\n", newline)
         if document.content.startswith(b"\xef\xbb\xbf"):
             serialized = "\ufeff" + serialized
         if source_text.endswith(("\n", "\r")):
--- a/engine_adapters/writeback.py
+++ b/engine_adapters/writeback.py
@@ -1,14 +1,16 @@
 """Common validation and rendering for declarative adapter writeback plans.
 
-Adapters describe safe text-span operations.  This module re-checks the plan
-against the live source snapshot and renders changed lines in memory.  It has
-no filesystem write authority; workflow code remains responsible for path
-resolution, transaction journaling, and atomic writes.
+Adapters describe safe text-span or semantic JSON catalog operations. This
+module re-checks the plan against the live source snapshot and renders changed
+files in memory. It has no filesystem write authority; workflow code remains
+responsible for path resolution, transaction journaling, and atomic writes.
 """
 
 from __future__ import annotations
 
+import copy
 import hashlib
+import json
 import ntpath
 from typing import Mapping, NoReturn, Sequence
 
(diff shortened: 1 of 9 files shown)


engine_adapters/writeback.py
```diff
@@ -249,11 +339,34 @@ def render_writeback_plan(
...
         lines[operation.line] = (
             lines[operation.line][: operation.start_col]
             + operation.replacement_fragment
             + lines[operation.line][operation.end_col :]
         )
+    for rel_path, data in json_documents.items():
+        document = json_source_documents[rel_path]
+        source_text = document.text()
+        newline = "\r\n" if "\r\n" in source_text else "\n"
+        serialized = json.dumps(data, ensure_ascii=False, indent=2)
```
hu-border-collie, line 366: **[bug]** Catalog rendering does not preserve CRLF. `json.dumps(..., indent=2)` always writes `\n`. The detected `newline` is applied only as a possible trailing terminator, so a CRLF source becomes mixed endings (`{\n  ... }\r\n`). I reproduced this with the same BOM+CRLF fixture the test builds: 9 interior LFs and a single trailing CRLF. JSON.parse will still accept the file, but this breaks the PR/docs invariant ("preserve CRLF"), dirties Windows catalogs, and makes the next snapshot hash a formatting rewrite rather than a translation update.

**Suggestion:** After dumps, normalize interior newlines to the detected terminator before adding a trailing newline, e.g. `serialized = json.dumps(...); if newline != "\n": serialized = serialized.replace("\n", newline)`, then prepend BOM and the trailing terminator. Keep `atomic_write_many_lines(..., newline="\n")` so already-terminated lines are not translated again.

tests/test_engine_adapter_p5_tyrano.py
```diff
@@ -263,16 +269,247 @@ def test_extract_occurrences_only_returns_extractable_candidates(self):
...
+                (translated,),
+                snapshot.project.source_documents,
+            )
+            rendered = "".join(
+                render_writeback_plan(plan, snapshot.project.source_documents)[
+                    "data/others/lang/ch.json"
+                ]
+            )
+            self.assertTrue(rendered.startswith("\ufeff"))
+            self.assertIn("\r\n", rendered)
```
hu-border-collie, line 507: **[suggestion]** `test_catalog_writeback_preserves_utf8_bom_and_crlf` only checks `startswith("\ufeff")` and `"\r\n" in rendered`. That passes for mixed endings and does not lock JSON key order, which the PR also claims to preserve. A future `sort_keys=True` or another dumps option would not fail this test.

**Suggestion:** Assert there is no bare LF (`"\n" not in rendered.replace("\r\n", "")`), that every split line uses the original terminator, and that `list(json.loads(rendered).keys())` (and nested scene/tag object keys) match the pre-write catalog order.

engine_adapters/writeback.py
hu-border-collie: Fixed in fa324a7: JSON indentation newlines are now normalized to the source catalog's detected CRLF convention before BOM and trailing-newline handling.

tests/test_engine_adapter_p5_tyrano.py
hu-border-collie: Addressed in fa324a7: the regression test now rejects bare LF characters, compares top-level and nested key order, and verifies the updated translation value.

